PR #69, opened 2014-03-16
Outputs tests with no callbacks as a TODO test.

aychtang:
Patching #62, referencing any defined tests with no callback in the TAP output as a TODO test, similar to if the TODO directive described in TAP spec is used (https://metacpan.org/pod/release/PETDANCE/Test-Harness-2.64/lib/Test/Harness/TAP.pod#TODO-tests).

### DIFF
--- a/example/no_callback.js
+++ b/example/no_callback.js
@@ -1,0 +1,3 @@
+var test = require('../');
+
+test('No cb test');
--- a/lib/test.js
+++ b/lib/test.js
@@ -178,6 +178,9 @@ Test.prototype.end = function (err) {
 
 Test.prototype._end = function (err) {
     var self = this;
+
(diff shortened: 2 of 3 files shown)


aychtang:
Noticed this was still open, not sure who to check with on this. @Raynos?

Tests without a cb still are still silent today. If you want it implemented as a TODO / SKIP directive let me know and I can fix the conflict. Otherwise you can close the PR.

ronkorving:
FWIW, I've implemented TODO as a proper option to run a test, but allow it to fail (as the TAP spec indicates) in #254

ljharb:
@aychtang could you check the "allow edits" checkbox on the RHS of the PR?

aychtang:
@ljharb yep, all done

PR #74, opened 2014-03-25
Replace a and b refs in readme with actual and expected

bclinkinbeard:
I can never remember the order, and figured I might not be alone.


domenic:
Seems to have been merged at some point

PR #94, opened 2014-07-19
Comments should not make exit code incorrect. Fixes #92

### LINKED ISSUE #92: t.comment() = fail exit code
Is this correct behaviour?

Including any test comments seems to fail a suite.

```js
// test.js
var test = require('tape')

test('comment test', function(t) {
  t.comment('this is a comment')
  t.ok(true)
  t.end()
})
```

```
> tape test.js
TAP version 13
# comment test
# this is a comment
ok 1 (unnamed assert)

1..1
# tests 1
# pass  1

# ok


> echo $?
1
```

tmcw:


grit96:
This should be reopened because this patch was never merged in and it is still a bug #92

PR #99, opened 2014-08-15
Added t.diag() method to print yaml metadata

### LINKED ISSUE #91: Could use a diag() Method
It's pretty typical for TAP implementations to offer an interface for emitting diagnostic messages as TAP comments. Examples:
- [Test::More](https://metacpan.org/pod/Test::More#Diagnostics) (Perl)
- [pgTAP](http://pgtap.org/documentation.html#diagnostics) (Postgres)
- [Test.More](http://openjsan.org/doc/t/th/theory/Test/Simple/0.29/lib/Test/More.html#Diagnostics) (JavaScript)

I tried to figure out where to plug one in to submit a patch, but the path to the output handle was over my head, so I used `console.log()` for my immediate use. Would be nice to have a canonical place for it.

andrewdeandrade:
This pull requests implements a `t.diag()` method that takes JSON and can print out the json as a YAML block. FWIW this adds jeremyfa/yaml.js as a new dependency, but that dependency works in browsers.

This would close #91 but before we do, I think it may be worth discussing this a bit.

The [TAP specification](http://testanything.org/tap-version-13-specification.html) states:

```
After a test line a block of lines starting with '---' and ending with '...' will 
be interpreted as an inline YAML document providing extended 
diagnostic information about the preceding test.
```

This seems to imply that extended diagnostic information should be attached to the preceding test. Having a standalone `diag()` method in a way implies that the diagnostic information to be printed is standalone information.

Now, the way I see it, there are three options here:
- Keep this pull request the way it is. The downside is that the diagnostic information is not "attached" to the assertion in question. The upside is that this frees someone up to print that diagnostic information later, but before the next assertion is executed. The reasons for this may be that the diagnostic information is not available at the time the assertion is executed. 
- Permit someone to add diagnostic information as a third argument, so that you could do `t.ok(true, 'Some message', someDiagnosticJSONObj)`. However, his approach would require quite a bit of reworking of `tape` since tape uses the third argument internally as an `extra` argument. This means that you'd have to do some whitelisting of the `extra` property object when you print it out.
- Allow both approaches.

Lastly, should we allow people to pass in YAML content directly into the `diag` method? AFAICT, the only thing necessary to do this is to check if the argument to `diag()` is a string or a plain old javascript object, and handle it accordingly.

Anyone have any strong thoughts here?

(FWIW I also looked into using `YAML.stringify` in the [`errorYAML` function](https://github.com/malandrew/tape/blob/87a882c83459c065e299fee1f27c31097897992c/lib/results.js#L149), but that didn't really work because it would YAMLize the `actual` and `expected` properties, even if they were JSON objects.)


Raynos:
Do we still want this ? If so we can #shipit

vvo:
@Raynos Yes! tap-parser supports diag so it would be good

gabrielcsapo:
👍 can we get this merged in?

ljharb:
It needs a rebase and conflicts resolved, or, the "allow edits" checkbox checked.

gabrielcsapo:
@isaacs are you maintaining this repo also?

ljharb:
@gabrielcsapo please don't ping arbitrary people unnecessarily; this PR can't be merged until the OP, @malandrew, updates it.

gabrielcsapo:
I was going to rebase the commit if @malandrew didn't respond. I was asking the question to `Isaacs` if this is being maintained any longer or if I should just move to `tap`. @ljharb

ljharb:
Yes, of course tape is being maintained, by multiple people.

I'd prefer the original poster be the one to do it, so we can modify this PR in-place instead of creating another one and polluting the git log.

isaacs:
I am not maintaining tape.  All the tap-related things I do are under the `tapjs` github org.

gabrielcsapo:
@isaacs does that mean node-tape is being deprecated in favor of tapjs?

isaacs:
@gabrielcsapo No.  Tape and node-tap are separate projects.  Always have been.

gabrielcsapo:
@isaacs okay, but the output is based on the tap standard version 13 currently right? I was looking to see if tape would be updated to adhere to that spec.

ljharb:
@andrewdeandrade can you please check the "allow edits" checkbox on the RHS of the PR?

lastthyme (approved):

Raynos:
Closing out due to lack of activity.

ljharb:
I prefer to keep inactive PRs open, in case the author returns.

PR #103, opened 2014-09-15
Do not intercept exceptions

Raynos:
It's weird that tape is intercepting exceptions and will report them as TAP error.

Exceptions are actually bugs in the test code and should not output TAP failures but just crash the test program instead.

This also makes it easier to test and debug tests themself as you can just use standard tools, like any standard uncaught exception tooling or any standard debugging thrown exception tooling.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -69,14 +69,7 @@ Test.prototype.run = function () {
         return this._end();
     }
     this.emit('prerun');
-    try {
-        this._cb(this);
-    }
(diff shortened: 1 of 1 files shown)


ehmicky:
For users trying to report uncaught exceptions as failed tests, I've created [a library](https://github.com/ehmicky/log-process-errors) that does just that. More information [here](https://github.com/substack/tape/issues/106#issuecomment-478609460).

PR #124, opened 2014-12-04
Make `tape` behave more like node `assert`.

Raynos:
This removes excess methods on the `Test` prototype.

There are a bunch of methods and aliases on the `Test` prototype
   that are not defined on assert for node core.

This adds a bunch of undocumented complexity and unnecessary
   deviation in the way that `tape` is used.

I'd be nice to remove these aliases and make `tape` a drop in
   replacement for `assert`.

This means that tape tests are more consistent and it means 
    it's easier to find the documentation by just reading the
    the node assert documentation.

This is a breaking change. Recommended version **major**.

cc @substack.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -112,7 +112,7 @@ Test.prototype.end = function (err) {
     }
     
     if (this.calledEnd) {
-        this.fail('.end() called twice');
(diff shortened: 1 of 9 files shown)


Raynos:
This ship has long sailed.

PR #127, opened 2014-12-14
Remove uncaught-exception listener

Raynos:
A testing library should not have an uncaught-exception listener, it makes testing hard code like uncaught exceptions harder.

For example
`
```
$ node endpoints/user/test/
TAP version 13
# can register

/home/raynos/projects/manga-feed/node_modules/tape/index.js:75
        throw err
              ^
RangeError: Maximum call stack size exceeded
```

Tape intercepts my range error and ruins the stack by rethrowing.

The particular PIPE bug we were gaurding against is better dealt with by adding 
an error listener on `process.stdout` or something.

Not really sure which EPIPE we are gaurding against. I recommend removing it
for now and we can add it back later.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,8 +64,20 @@ function createExitHarness (conf) {
     if (conf.exit === false) return harness;
     if (!canEmitExit || !canExit) return harness;
 
+    var inErrorState = false;
+
+    var $_fatalException = process._fatalException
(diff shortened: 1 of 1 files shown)


Raynos:
If no one objects to this change then I will land it.

Raynos:
cc @substack et al.

PR #130, opened 2014-12-31
Add a Function.constructor check to throws.

wilmoore:
Allows you to also write:

```
t.throws(fn, TypeError);
```

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -18,7 +18,7 @@ var getTestArgs = function (name_, opts_, cb_) {
     var name = '(anonymous)';
     var opts = {};
     var cb;
-    
+
     for (var i = 0; i < arguments.length; i++) {
         var arg = arguments[i];
         var t = typeof arg;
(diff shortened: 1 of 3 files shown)


lib/test.js
```diff
@@ -417,6 +418,11 @@ Test.prototype['throws'] = function (fn, expected, msg, extra) {
         expected = String(expected);
     }
 
+    if (expected instanceof Function) {
+        passed = caught.error.constructor === expected;
```
wilmoore: > `caught.error instanceof expected`

Are we 100% sure that we want to allow lose inheritance-based matching here? For example, if a `TypeError` is thrown, if I assert the type is `Error`, that would pass.

I can see cases where I'd want either of those behaviors (though, for my current use-case, I need the strict test). Perhaps we should consider a completely separate assertion for one or the other.

What do you think @Raynos?

Raynos: @wilmoore that's the semantics of `require('assert')` in node core on purpose.

Having different semantics from `require('assert')` is confusing and not worth it, even if they are "technically" better.

wilmoore: > Having different semantics from require('assert') is confusing and not worth it...

Indeed. I agree with that. Update coming.

Raynos:
lgtm :)

Issue #129, opened 2014-12-30
Intend output of subtests


binarykitchen:
I think that would be neat. At the moment this code

```js
test('middleware:', function(t) {
    ...

    t.test('with baiduspider', function(t) {
        ...
    })
})
```

outputs

```
...
TAP version 13
# middleware:
# with baiduspider
...
```

but indentation like that would be cool:

```
...
TAP version 13
# middleware:
    # with baiduspider
...
```

Can do a PR if you're too busy.

koresar:
You probably should do it.

Raynos:
Sounds good.

However lets do '# four spaces' instead of 'four spaces #'

koresar:
I hope nobody would mind if I do it? Thanks. :)

binarykitchen:
@koresar go ahead, i'm too busy with another crazy project of mine ...

koresar:
Please, see https://github.com/substack/tape/pull/136

substack:
Instead of inventing something we should wait to see what TAP14 decides: https://github.com/TestAnything/Specification/issues/2

DeusProx:
So 3 years later still no official answer? I would like to use this implementation but having no indentation for my subtests looks really unstructured.

PR #136, opened 2015-01-20
Subtests output indentation.

koresar:
Implements #129. The output would look like this. Not so pretty IMO.
```
TAP version 13
# Test name
# LEVEL0
ok 1 LEVEL0
#     Sub-test name
#     LEVEL1
    ok 2 LEVEL1
#         Sub-sub-test name
#         LEVEL2
        ok 3 LEVEL2
```

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -92,15 +92,15 @@ Results.prototype._watch = function (t) {
     var self = this;
     var write = function (s) { self._stream.queue(s) };
     t.once('prerun', function () {
-        write('# ' + t.name + '\n');
+        write('# ' + getIndentation(t._level) + t.name + '\n');
     });
     
     t.on('result', function (res) {
         if (typeof res === 'string') {
-            write('# ' + res + '\n');
+            write('# ' + getIndentation(t._level) + res + '\n');
             return;
         }
-        write(encodeResult(res, self.count + 1));
+        write(getIndentation(t._level) + encodeResult(res, self.count + 1));
         self.count ++;
 
         if (res.ok) self.pass ++
@@ -124,6 +124,12 @@ Results.prototype.close = function () {
 
     self._stream.queue(null);
 };
+
+function getIndentation(level) {
+  var indent = '';
+  while (level-- > 0) indent += '    ';
+  return indent;
+}
 
 function encodeResult (res, count) {
     var output = '';
--- a/lib/test.js
+++ b/lib/test.js
@@ -46,6 +46,7 @@ function Test (name_, opts_, cb_) {
     this.assertCount = 0;
     this.pendingCount = 0;
     this._skip = args.opts.skip || false;
+    this._level = args.opts.level || 0;
     this._plan = undefined;
     this._cb = args.cb;
     this._progeny = [];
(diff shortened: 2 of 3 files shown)


Raynos:
Indenting ok breaks tap. Try 'ok number 4 spaces text'

Raynos:
Also. Really this shouldn't be in tape. We should do this in tap-spec instead

scottcorgan:
Is there any indication in tap that the test description is from a sub test?

Raynos:
I don't think so. We would need tape specific heuristics. We could change the name output to be '# level 1 level 2 level 3'

Then you can detect sub tests with subset operations on test names.

scottcorgan:
That could work. I started doing something like that with https://github.com/scottcorgan/tessed to try and solve this very issue.

It would help a lot of people to have nice output for nested tests.

Perhaps an option to for a separator between test names for sub tests would be helpful `{sep: '-'}` ... something like that.

koresar:
I ok guys. For now I'll change tape to tap-spec.

Meanwhile, we got to decide on the nested output format. I'm all ears.

koresar:
Re the tap over tape. Unluckily I can't receive the output using the common harness:
```js
  var test = tape.createHarness();
  var tc = tap.createConsumer();
  test.createStream().pipe(tc);
```
because tap strips the leading hash characters and whitespaces. This means I can't just simply
> do this in tap-spec instead

as I am trying to test those whitespaces and hashes. :)

Subsequently the
> separator between test names for sub tests

is more acceptable solution because it's testable via tap. I have it working locally.
```
TAP version 13
# Output indentation
# LEVEL0 comment
ok 1  LEVEL0 assertion
# Output indentation -> Sub-test name
# Output indentation -> LEVEL1 comment
ok 2 Output indentation ->  LEVEL1 assertion
# Output indentation -> Sub-test name -> Sub-sub-test name
# Output indentation -> Sub-test name -> LEVEL2 comment
ok 3 Output indentation -> Sub-test name ->  LEVEL2 assertion
```
*Although this looks quite confusing to me.*

See the code which generates the output above:
```js
test('Output indentation', function(t) {
  t.comment('LEVEL0 comment');
  t.ok(true, 'LEVEL0 assertion');
  t.test('Sub-test name', function(tt) {
    tt.comment('LEVEL1 comment');
    tt.ok(true, 'LEVEL1 assertion');
    tt.test('Sub-sub-test name', function(ttt) {
      ttt.comment('LEVEL2 comment');
      ttt.ok(true, 'LEVEL2 assertion');
      ttt.end();
      tt.end();
      t.end();
    });
  });
});

```

koresar:
@Raynos @substack hello gentlemen.

Subtest output indentation. People need this. It's time to decide how you see it implemented.

Would you please express an opinion/desire/anything? What's a preferred output format?

Thanks

substack:
This is a protocol-level decision that shouldn't be up to tape. There's already active discussion for TAP14 about subtests: https://github.com/TestAnything/Specification/issues/2

When something lands in the spec, then we can support it. Until then, it's possible to consume tap output and reformat it however you want.

isaacs:
I commented on this: https://github.com/TestAnything/Specification/issues/2#issuecomment-76519391

I'm planning on implementing the `Test::More` style.  It's the one in common use.  Anything that lands in tap v14 will be either trivially convertible from Test::More style, or not worth using, imo.

I'm also refactoring node-tap quite a lot in the process, so it might be pretty easy to re-use some bits in tape, I'm not sure.

PR #142, opened 2015-01-30
Fixing backwards t.end explanation

TehShrike:
t.end asserts that the "err" object is not truthy https://github.com/substack/tape/blob/master/lib/test.js#L323


Raynos:
Thanks :)

PR #149, opened 2015-03-10
Only check for errors in callback if exist

Raynos:
This removes an unnecessary assert null in
case you do `thing.close(assert.end)` and
the `close()` operation does not have an
error.

This keeps the TAP output cleaner.

reviewers: @jcorbin @malandrew 

cc: @substack


Raynos:
Also added @malandrew as reviewer.

domenic:
LGTM

PR #152, opened 2015-04-02
Update deps + add readme section on things that go well with tape

domenic:
Looking to do a patch release after this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 node_js:
-  - "0.8"
   - "0.10"
+  - "0.12"
+  - "iojs"
--- a/package.json
+++ b/package.json
@@ -9,17 +9,17 @@
     "test": "test"
   },
   "dependencies": {
-    "deep-equal": "~0.2.0",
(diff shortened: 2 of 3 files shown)


domenic:
It seems our deps have switched to using carets, so the build fails on Node 0.8.

A few options:

- Only update to deps that don't use carets. (That'll take care of object-inspect, which is the most important one to me.)
- Update travis.yml to install a new npm before testing, and don't bump the major version because people on Node 0.8 really need to update npm by now.
- Update travis.yml to install a new npm before testing, but bump the major version since this is a new requirement and could break builds
- Drop 0.8 support and bump the major version.

Since I want to get this done soon I'd guess the first one is going to be most acceptable to people. Confirm/deny?

Raynos:
I'm :+1: for dropping 0.8.

Raynos:
Once you fix travis; by any means. Then #shipit

substack:
You can just add:

```
before_install:
  - npm install -g npm@2.7.5
```

to `.travis.yml`. It's not 0.8 as such, just that travis ships with an ancient version of npm for node 0.8.

PR #153, opened 2015-05-09
Clarify comment documentation

davidtheclark:
I'm getting that `test.comment()` is undefined; so I'm thinking that `t.comment()` is what was intended to be documented. Right?

I moved it only so that it is next to the other `t` methods.


Raynos:
Thanks!

PR #156, opened 2015-06-03
Add info about bin usage to readme.

jmm:
The bin appears to be totally undocumented.

Please review for accuracy. [Rendered](https://github.com/jmm/tape/blob/bin-usage/readme.markdown).


Raynos:
Thanks!

PR #163, opened 2015-07-14
Cache `Object#hasOwnProperty` in case clients break it as part of tests

ljharb:
Obviates https://github.com/substack/object-inspect/pull/2.

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -3,6 +3,7 @@ var inherits = require('inherits');
 var through = require('through');
 var resumer = require('resumer');
 var inspect = require('object-inspect');
+var hasOwn = Object.prototype.hasOwnProperty;
 var nextTick = typeof setImmediate !== 'undefined'
     ? setImmediate
     : process.nextTick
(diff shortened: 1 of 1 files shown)


lib/results.js
```diff
@@ -3,6 +3,7 @@ var inherits = require('inherits');
 var through = require('through');
 var resumer = require('resumer');
 var inspect = require('object-inspect');
+var has = Object.prototype.hasOwnProperty;
```
ljharb: If `tape` targets ES5 (rather than ES3), or, if we want to add `function-bind` as a dep, then this could be `var has = Function.call.bind(Object.prototype.hasOwnProperty);` or `var has = require('function-bind')(Function.call, Object.prototype.hasOwnProperty);` and the function at the bottom could be removed entirely.

Raynos: No bind needed

Raynos:
Lgtm.

lib/results.js
ljharb: The benefit of adding the bind is that `Function.prototype.call` could be removed/broken, and this module wouldn't break. It's more robustness than most bother with, though :-)

vvo: Would be cool to keep tape ES3 compliant :)

ljharb: Agreed! The `function-bind` dependency would preserve that.

PR #165, opened 2015-07-19
Update `object-inspect`

### LINKED ISSUE #158: Attempting to display Symbols crashes in node 0.12
`t.equal(Symbol(), false)` will crash on node 0.12, because `String(Symbol())` incorrectly throws there.

If https://github.com/substack/object-inspect/pull/1 is merged and published, and bumped here, it will fix this problem.

ljharb:
This fixes #158 by ensuring that a minimum of v1.0.1 of `object-inspect` is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "defined": "~0.0.0",
     "glob": "~5.0.3",
     "inherits": "~2.0.1",
-    "object-inspect": "~1.0.0",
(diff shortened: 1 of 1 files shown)


substack:
The version is already floating, so I don't think this is necessary.

ljharb:
Fair. This would just ensure that nobody can use a broken version of `object-inspect`

PR #166, opened 2015-07-21
Added --save-dev to install instruction

mathiasrw:
My 2 cents...

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -344,7 +344,7 @@ $ node object.js test/x.js test/y.js
 With [npm](https://npmjs.org) do:
 
 ```
-npm install tape
+npm install tape --save-dev
 ```
 
 # license


troutowicz:
Not everyone that installs tape plans on developing tape.

substack:
@troutowicz: this would install tape as a devDependency, not a dependency in package.json. I think that is what most people will want.

ljharb:
Although I personally don't use the save options on `npm install`, since it messes with my package.json formatting, I think including them as a default is the appropriate thing to do for the majority of users. This LGTM.

mathiasrw:
:+1:

wavded:
:+1:

troutowicz:
> this would install tape as a devDependency

Oops... :sob:

PR #170, opened 2015-08-08
Expose the main test harness

### LINKED ISSUE #148: What's the best way to report pass / fail to sauce labs?
Orchestrating sauce labs runs with `selenium-sauce`, what's the best way to report pass / fail to Sauce Labs?

```
var
  test = require('tape'),
  SeSauce = require('selenium-sauce');

// Loads the config file and invokes the callback once for each browser
new SeSauce(
    {   // Configuration options
        quiet: false,           // Silences the console output
        webdriver: {            // Options for Selenium WebDriver (WebdriverIO)
            user: process.env.SAUCE_USERNAME,
            key: process.env.SAUCE_ACCESS_KEY,
            desiredCapabilities: [{
                browserName: 'chrome',
                version: '27',
                platform: 'XP',
                tags: ['examples'],
                name: 'This is an example test'
            }]
        },
        httpServer: {           // Options for local http server (npmjs.org/package/http-server)
            port: 8080              // Non-standard option; it is passed into the httpServer.listen() method
        },
        sauceLabs: {            // Options for SauceLabs API wrapper (npmjs.org/package/saucelabs)
            username: process.env.SAUCE_USERNAME,
            password: process.env.SAUCE_ACCESS_KEY
        },
        sauceConnect: {         // Options for SauceLabs Connect (npmjs.org/package/sauce-connect-launcher)
            username: process.env.SAUCE_USERNAME,
            accessKey: process.env.SAUCE_ACCESS_KEY
        },
        selenium: {             // Options for Selenium Standalone Server (npmjs.org/package/selenium-standalone). Only used if you need Selenium running locally.
            args: []                // options to pass to `java -jar selenium-server-standalone-X.XX.X.jar`
        }
    },
    function(browser) {

        // Initialize the browser
        browser.init(function(err) {
            if(err) throw err;

            // Load a url into the browser
            browser.url('http://localhost:8080/test.html', function() {
                var result;
                // Tell SauceLabs that the test was successful
                result = // what do we put here?
                browser.passed(result, myCompletionCallback);
            });
        });
    }
);
```

TehShrike:
Allows reporting failed/succeeded tests to SauceLabs/BrowserStack.

Fixes #148

This fix was even simpler than I'd imagined, so I'm opening this pull request instead of publishing the fork right away.


index.js
```diff
@@ -35,9 +35,13 @@ exports = module.exports = (function () {
         }
         return harness.createStream(opts);
     };
-    
+
+    lazyLoad.results = function () {
```
Raynos: this is not general enough; let's just `lazyLoad.getHarness = getHarness;`

Raynos:
Will merge after feedback addressed.

TehShrike:
Sounds good to me!  New commit pushed up.

Raynos:
Published v4.1.0

PR #173, opened 2015-08-12
Adding --init option to tape cli

joakimbeng:
With this change one can run:

```sh
$ tape --init
```

To add `tape` to the current project, including a npm test script.

What do you think?

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -2,6 +2,13 @@
 
 var path = require('path');
 var glob = require('glob');
+var init = require('tape-init');
+var hasFlag = require('has-flag');
+
+if (hasFlag('init')) {
+  init();
+  return;
+}
 
 process.argv.slice(2).forEach(function (arg) {
     glob(arg, function (err, files) {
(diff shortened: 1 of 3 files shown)


ljharb:
What if i already have a "test" script specified? The default seems to be `test/**/*.js` - that's a convention, and there's no guarantee any given `tape` user follows that pattern.

It seems like this is a convention that perhaps shouldn't be bundled into `tape`?

joakimbeng:
@ljharb if a "test" script is already specified there are three scenarios:

1. it contains the `npm init` default: it will be _replaced_
2. it contains the string "tape" (i.e. a test script containing "blue-test ..." will also match): it's left _untouched_
3. it contains something else, e.g. "xo": the content will be _appended_ with " && tape ..." (if one is using another test runner and don't want this, one could wonder why they did run this in the first place?)

About the pattern.. Yes, that's a convention (though a quite common one). How would you like it if there were no default and that one must specify it on init, for instance: `tape --init 'tests/*.js'`?

Raynos:
:-1: tape is not a scaffolder. 

Configuring your package.json is the responsibility of a project scaffolder.

The fact that tape even has a globbing cli is massive feature creep already.

joakimbeng:
I understand, no worries. I'm closing this then.

PR #181, opened 2015-08-23
Fix Issue #180: t.throws() giving false positive regexp matches

### LINKED ISSUE #180: t.throws giving false positive RegExp matches
Hey,

Thanks for writing tape! 

I believe I found a bug in `test.throws`, that should have been failing much earlier. The problem is illustrated by the following example (which you can run in Chrome, e.g.):

    > var re = /foo bar/;
    undefined
    > re.test(new Error('foo bar'));
    true
    > re = /^foo bar$/;
    /^foo bar$/
    > re.test(new Error('foo bar'));
    false

Unfortunately, that's what `t.throws` is doing under the hood. This is causing tests like the following to fail

    test('does not match regex', t => {
      let fn = () => {
        throw new Error('foobar bazqux')
      };
    
      t.throws(fn, /^foobar bazqux$/)
    });

as well as [this test](https://github.com/yangmillstheory/tape/commit/76fe01db85d2052d30f32b2a02d015deb909c585).

yangmillstheory:


yangmillstheory:
Fixes [#180](https://github.com/substack/tape/issues/180).

lib/test.js
```diff
@@ -434,16 +434,15 @@ Test.prototype['throws'] = function (fn, expected, msg, extra) {
         fn();
     } catch (err) {
         caught = { error : err };
-        var message = err.message;
-        delete err.message;
-        err.message = message;
```
yangmillstheory: I removed these lines, I don't think they actually do anything?

yangmillstheory:
`npm test` is passing.

lib/test.js
ljharb: The impact they have is, if an error had a getter as its "message" property, this converts it to a straight assignment.

However, if these are important, there should be a test that ensures they're not removed.

ljharb:
It seems to me that this isn't actually a bug, it's desired behavior. The stringification of the Error object contains much more information than just the message, and the regex is intended to match against all of that.

lib/test.js
yangmillstheory: Ah, I see the intended effect.

yangmillstheory:
Thanks @ljharb, I see your point.

    > String(new TypeError("foobar bazqux"))
   <- "TypeError: foobar bazqux"

Is it the case the that only new information contained here will be the constructor name? Although, I can imagine people providing `toString()` implementations for custom errors. 

This does have some overlap with `t.throws` accepting a `Function`, I think.

yangmillstheory:
Closing this, as I don't see much point in being more restrictive with regex matches - it's a breaking change and going forward, it remove support for matching errors for people who would export custom exceptions with their own `toString()`.

PR #191, opened 2015-09-17
notDeepLoosEqual to set .notExpected property instead of .expected

hville:
Unlike `t.notEqual` and `t.notDeepEqual`, `t.notDeepLooseEqual` sets the reference value in the `.expected` property instead of the `.notExpected` property.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -417,7 +417,7 @@ Test.prototype.notDeepLooseEqual
         message : defined(msg, 'should be equivalent'),
         operator : 'notDeepLooseEqual',
         actual : a,
-        expected : b,
+        notExpected : b,
         extra : extra
     });
 };


ljharb:
This would be a breaking change, and it has no tests.

If this change isn't causing any tests to fail, tests will need to be added - otherwise, they'd need to be modified.

hville:
The PR passes all of the current tests and is consistent with all other methods. 

Creating a new test for this would invlove also creating new tests for `deepEqual`, `notDeepEqual`, `deepLooseEqual` and `notDeepLooseEqual` since the `.expected` and `.notExpected` properties are not validated in any of the current tests. This would be a seperate PR with a much wider scope.

ljharb:
It's not solely up to me, but I'd rather refuse to merge a PR that doesn't come with at least a regression test. Even a single test case, that breaks without your change, would be a start. If tests are passing without your change - and your change is necessary - then the lack of a failing test is part of the same bug.

hville:
Ok. Let me think about it because I don't really know what to test apart from the property itself. (I'm using these properties to pretty-format the output from the native result object instead of the text output)

Raynos:
lgtm.

hville:
Check the TAP spec and node's ```assert``` to compare and devise relevant tests and found that node's ```assert``` only uses the ```.expected``` property for all assertion types instead of ```.expected``` and ```.notExpected```. For consistency I would be better to change ```tape```'s properties accordingly (the opposite of the fix proposed in this PR) which could be a breaking change. I'm now using ```assert``` directly instead of jugling arround with ```tape```'s stream so this PR can be dropped and closed.

ljharb:
This seems subsumed by #454.

PR #195, opened 2015-09-28
Travis: Add Node v4

wbinnssmith:
I imagine the entry "4" will continue to be the latest stable release of the 4.x line on travis.


ljharb:
I don't see any reason to remove `iojs` if it continues to pass. Also, instead of `4`, you can use `node` to test the latest version of node.

wbinnssmith:
Hm. When node hits 5 (which it will in the coming months iirc), "node" will simply move onto 5 (unless travis decides to keep node pointing to the latest LTS instead of the latest node) and 4 will become untested. Either way, I imagine it will be best to be explicit. [See browserify's patch](https://github.com/substack/node-browserify/commit/97a122e9c885158733e0fa2f4eecf5d238202381).

ljharb:
¯\_(ツ)_/¯ not sure what node LTS has to do with an arbitrary library's chosen support level.

I don't mind putting `4` in here - not `4.0.0`, though - but I still think it's useful to keep `iojs`.

wavded:
I second that its still useful to keep `iojs`.

wbinnssmith:
All right, the patch simply adds "4" now.

ljharb:
LGTM

PR #200, opened 2015-10-15
update dependencies and tests (WIP)

chicoxyzzy:
WIP here is because I can't figure out what should I use instead of `tap.createConsumer` which is missing in node-tap >= 1.0.0. @isaacs is there any alternatives of `createConsumer` method in node-tap 2.2.x?

These tests should be fixed to work with tap 2.1.1
- [ ] array
- [ ] end-as-callback
- [ ] exit
- [ ] fail
- [ ] nested-sync-noplan-noend
- [ ] nested
- [x] only
- [x] timeoutAfter
- [x] too_many

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,10 @@
     "through": "~2.3.4"
   },
   "devDependencies": {
-    "concat-stream": "~1.5.0",
+    "concat-stream": "~1.5.1",
     "falafel": "~1.2.0",
     "js-yaml": "^3.3.1",
-    "tap": "~0.7.1",
+    "tap": "~2.1.1",
     "tap-parser": "^1.1.6"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


ljharb:
Why do `concat-stream` and `falafel` need updates?

isaacs:
node-tap just uses `tap-parser` to parse streams, it doesn't have a separate "consumer" concept any more.

chicoxyzzy:
@ljharb actually I'm trying to update tap to 2.1.1 as well but I am still tying to figure out what should I do to update tests for:
- array
- end-as-callback
- exit
- fail
- nested-sync-noplan-noend
- nested
- only
- timeoutAfter
- too_many

chicoxyzzy:
I don't work on this for almost a year so I'll better close it

ljharb:
Thanks - this was done by #318 recently anyways.

Issue #102, opened 2014-09-08
ES6 for tape globbing util


emilbayes:
Right now it's not possible to enable `--harmony` when using the tape globbing util.
I know that passing arguments to the shebang is not standardised and not possible across OS's. Is there some other way around this?

saulshanabrook:
@Raynos is it possible to have this command as an npm script without referring directly to the path? Something like:

```json
  "scripts": {
    "test": "node --harmony tape 'tests/*.js'"
  },
```

But that doesn't work, because it cannot find module tape.

The reason is that I didn't install tape in that directory, because I am [running through docker](http://www.saulshanabrook.com/npm-docker-sharing-volumes/)...

PR #203, opened 2015-10-24
Implement shebang using babel-node

evcohen:
Add babel-node functionality to running tape from command line. This
takes advantage of the glob util and handles important es6 features not
covered by —harmony flag (as suggested in https://github.com/substack/tape/issues/102).

### DIFF
--- a/bin/babel-tape
+++ b/bin/babel-tape
@@ -1,0 +1,14 @@
+#!/usr/bin/env babel-node
+
+var path = require('path');
+var glob = require('glob');
+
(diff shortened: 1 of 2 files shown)


wavded:
How is this different from https://github.com/wavded/babel-tape-runner ?

evcohen:
The difference is that I didn't know that existed :) We should put that in the docs. Thanks @wavded !

yoshuawuyts:
Or alternatively:
```sh
$ node -r babel/register test
```

PR #210, opened 2015-11-06
Add tape-promise into 'other'

jprichardson:
Unlike [`blue-tape`](https://www.npmjs.com/package/blue-tape), [`tape-promise`](https://github.com/jprichardson/tape-promise) has the following benefits:
1. Doesn't explicitly depend upon `tape`. i.e. can directly modify `tape` if need be and not fixed to the `tape` version of `blue-tape`.
2. Explicitly supports async/await.
3. Has better documentation.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -127,7 +127,7 @@ By default, uncaught exceptions in your tests will not be intercepted, and will 
 ## other
 
 - CoffeeScript support with https://www.npmjs.com/package/coffeetape
-- Promise support with https://www.npmjs.com/package/blue-tape
+- Promise support with https://www.npmjs.com/package/blue-tape or https://www.npmjs.com/package/tape-promise
 - ES6 support with https://www.npmjs.com/package/babel-tape-runner or https://www.npmjs.com/package/buble-tape-runner
 - Different test syntax with https://github.com/pguth/flip-tape (warning: mutates String.prototype)
 


wavded:
@jprichardson why not patch blue-tape?  to me, the functionality seems so close.

1. `blue-tape` does support `async/await` (really its a babel thing)
2. docs haven't been a problem for me but a pull could fix that
3. i haven't experience an issue needing 1 (@spion has been good with pulling version bumps in), but maybe you could elaborate on why?  i kinda like having tape explicitly defined as a dep so i know it works and has been tested.

jprichardson:
1. While you're right, it doesn't explicitly (stated via Readme / include tests) support it.
2. Yes.
3. This is the most important aspect for me. I don't want it to depend upon Tape. I have a bunch of unreleased tape utilities that just decorate tape. What's cool is that people can leverage functional composition to pick/choose which they like.

e.g....

```js
var tape = require('tape')
var tapePromise = require('tape-promise')
var tapeSomething = require('tape-something')
var compose = require('some-library-that-has-compose-maybe-lodash')

var _test = compose(tapePromise, tapeSomething)
var test = _test(tape)

test('should do someting', function (t) { /** ... **/ }
```

ljharb:
fwiw, I don't think anything should explicitly state support for `async/await`, since that betrays a fundamental misunderstanding of what `async/await` is - it's nothing more than syntax sugar for nested promises, so *everything and anything* that works with promises supports `async/await`.

yoshuawuyts:
@ljharb that's the best description of `async/await` I've seen so far.

lijunle:
I could like to use `tape-promise` instead of `blue-tape`.

I am writing the following scripts to run ES6 test and ES6 coverage. It is using `tape` directly, I could like to declare `tape` as devDependencies and with `tape-promise`.

```js
"scripts": {
  "test": "babel-node node_modules/tape/bin/tape ./test/**/*.js",
  "test-coverage": "babel-node node_modules/isparta/bin/isparta cover node_modules/tape/bin/tape -- ./test/**/*.js"
}
```

ljharb:
@jprichardson could you check "allow edits" on the right hand side of this PR?

Issue #175, opened 2015-08-14
Emit event when all tests are done


tsironis:
I've been trying to run a mock API server that I wrote and use it with tape, but it seems that when the mock API server is running, ```tape``` cannot exit. Although I get the usual message,

```
1..39
# tests 39
# pass  39

# ok
```
the suite never really ends.

Is there a way to know when all tests are done? I've read the docs and couldn't find anything relevant.

webthusiast:
I have the same problem. My use case: I have a web page which runs tape in the browser, and I have a phantomjs-script that loads that page. When the tests are done (and all results have been written), I want to exit the phantomjs process.

I had a look at the code, and it looks like this isn't supported. In node, the process exits when the result-stream ends. But this stream is a closure variable, so not accessible outside the `createExitHarness` function.

Pajk:
:+1: 

This is my quick&dirty solution as the one @webthusiast mentioned did not work for me.

https://github.com/Pajk/tape/commit/a4355d8466a0894855c470164f9f7c2e5d227af5

It's called only when all tests pass but that's enough my case. I use it to disconnect node-postgres connection pool.

akiva:
+1 I, too, encounter this, and find that the usual solution of implementing something akin to:

```javascript
tape('teardown', function (t) {
  server.close(...);
});
```
does not actual exit the test script process and the console will remain as @tsironis indicates (meaning, the server I am trying to close does, in fact, close).

rhys-vdw:
Also interested in this feature. I tried @webthusiast's method, but the 'end' event never fires.

rhys-vdw:
Or perhaps my problem is different - there is no summary printed at the end of the tests... It looks like the tests are simply not ending.

Raynos:
+1

I've decided this is now a useful feature >_>;

rhys-vdw:
@Raynos How does tape know when to emit the end event if there are async tests? Do all calls to `test` need to occur synchronously?

Raynos:
At some point the program exits and prints the number of oks. Thats a good place to hook into.

Have to read more code.

I think it's simply reference counting.

rhys-vdw:
My [asyc tests](https://github.com/rhys-vdw/data-mapper/blob/73308b81b4a6d1538d89d5335ae733710ee7850c/tests/integration/mapper/retrieval.js) just hangs indefinitely.

akiva:
[...]

PR #211, opened 2015-11-13
add tape.tearDown handler

webthusiast:
Should fix issue #175 (though without actually emitting an event).


Raynos:
This seems reasonable.

PR #217, opened 2015-11-21
Removed unreachable Results.prototype.only code

taylorhakes:
I was looking though the results.js file and noticed invalid code.
```js
if (this._only) {
  self.count ++;
  self.fail ++;
  write('not ok ' + self.count + ' already called .only()\n');
}
```
`self` and `write` are not defined. I tried to write a failing test, but found out that the code was unreachable because there is a similar check in index.js before reaching this code.
```js
results.only(name);
```

This pull requests removes the broken unreachable code. I also added a test for duplicate `test.only` calls.

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -84,11 +84,6 @@ Results.prototype.push = function (t) {
 };
 
 Results.prototype.only = function (name) {
-    if (this._only) {
-        self.count ++;
-        self.fail ++;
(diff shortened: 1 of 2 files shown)


Raynos:
Thanks for this bugfix :)

Raynos:
Published v4.4.0

Issue #220, opened 2015-12-09
No line number provided in Windows on failed tests (and proposed fix)


AlexDM0:
Hi,

When I'm using windows with nodejs, tape does not provide an at: field with the line numbers if a test fails. 

It is fixed by placing:
```
var windowsPath = dir.substr(1,1) === ":"; // check for windows path if the second letter is a :
if (windowsPath)
  dir = dir.replace(/(\\)/g,"/").substr(2);  // swap \ for / and remove the D: or C: from C:\
```
at https://github.com/substack/tape/blob/master/lib/test.js#L212

and inserting:
```
if (windowsPath) 
  err[i] = err[i].replace(/(\\)/g,"/");  // swap \ for / 
```
before https://github.com/substack/tape/blob/master/lib/test.js#L214

This bug is caused by the drive letter before the path which you do not seem to expect in the rest of the logic and the inverted slashes.

I can do a pull if you'd like?

Regards

ljharb:
@AlexDM0 Given #316, is this issue resolved?

AlexDM0:
I think it can be closed, thanks!

PR #221, opened 2015-12-09
fixed windows path support for error messages

AlexDM0:
Might as well do the pull request for #220

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -209,8 +209,14 @@ Test.prototype._assert = function assert (ok, opts) {
         var e = new Error('exception');
         var err = (e.stack || '').split('\n');
         var dir = path.dirname(__dirname) + '/';
-        
+        var windowsPath = dir.substr(1,1) === ":"; // check for windows path if the second letter is a :
+        if (windowsPath)
+            dir = dir.replace(/(\\)/g,"/").substr(2);  // swap \ for / and remove the D: or C: from C:\
+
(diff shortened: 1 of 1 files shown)


ljharb:
Instead of detecting windows paths, can we just use the `path` module, which has OS-specific directory separators?

lib/test.js
```diff
@@ -209,8 +209,14 @@ Test.prototype._assert = function assert (ok, opts) {
         var e = new Error('exception');
         var err = (e.stack || '').split('\n');
         var dir = path.dirname(__dirname) + '/';
```
ljharb: specifically here - instead of using a hardcoded slash, we could use `path.sep`?

AlexDM0:
Or we can write 4 lines of code and avoid yet another dependency.....

ljharb:
a) more dependencies is always better than reinventing wheels.
b) path is built into node, and when browserified, is appropriately shimmed.

AlexDM0:
Sure, path is also fine. As long as the bug is fixed I suppose.

Raynos:
will merge if using portable `path` code.

I also dont have a way to verify or test this on windows itself.

hgwood:
I have given another try to this fix: #295.

ljharb:
@AlexDM0 Given #316, should this PR be closed, or rebased and re-evaluated?

AlexDM0:
I think it can be closed, thanks!

Issue #126, opened 2014-12-11
Tests never end


tommy31:
Can someone explain to me what i'm doing wrong? Because my test never end..

```Javascript
var express     = require('express');
var test	       = require('tape');

test('test', function (t) {
        var app;

	t.plan(1);

        function start() {
           t.pass('server started');
        }

        function error(err) {
           t.error(err, 'server startup failed');
        }

        app = express();

	app.listen('8080', 'localhost', function(err) {
	   if (err) return error(err)

	   return start();
	});
});
```

Outputs : 

> tape tests/*.js

```
TAP version 13
# test
ok 1 server started
```

PR #224, opened 2015-12-19
Add flag to require modules before running tests

mstade:
By adding the `--require` flag (and `-r` alias) we allow users to require modules before running their tests. An example of this might be adding source-map-support, like so:

```bash
$ tape -r source-map-support/register test/*.js
```

Another example might be to install the [babel require hook](http://babeljs.io/docs/usage/require/):

```bash
$ tape -r babel-register test/*.js
```

Another potential benefit is the ability to more easily run external test suites: 

```bash
$ npm install test-suite-rfc1337
$ tape -r test-suite-rfc1337
```

Modules are found using the node module resolution algorithm, so if one wants to load a local module it is necessary to prepend `./` or `../`, like so:

```bash
$ tape -r ./my/local/module
```

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -1,12 +1,32 @@
 #!/usr/bin/env node
 
-var path = require('path');
+var resolveModule = require('resolve').sync;
+var resolvePath = require('path').resolve;
+var parseOpts = require('minimist');
 var glob = require('glob');
 
-process.argv.slice(2).forEach(function (arg) {
+var opts = parseOpts(process.argv.slice(2), {
+        alias: { 'r': 'require' },
+        string: 'require',
+        default: {'r': [] }
+    });
+
+var cwd = process.cwd();
+
+/* If only one require is specified, the value of `opts.require`
+ * will be a string. This is why we concatenate.
+ */
+;[].concat(opts.require).forEach(function(module) {
+    /* The `module &&` ensures we ignore `-r ""`, trailing `-r` or other
+     * silly things the user might (inadvertedly) be doing.
+     */
+    module && require(resolveModule(module, { basedir: cwd }));
+});
+
+opts._.forEach(function (arg) {
     glob(arg, function (err, files) {
         files.forEach(function (file) {
-            require(path.resolve(process.cwd(), file));
+            require(resolvePath(cwd, file));
         });
     });
 });
--- a/bin/tape
+++ b/bin/tape
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
-require('source-map-support/register');
 
 var path = require('path');
 var glob = require('glob');
--- a/bin/tape
+++ b/bin/tape
@@ -18,10 +18,11 @@ if (typeof opts.require === 'string') {
 }
 
 opts.require.forEach(function(module) {
-    /* The `module &&` ensures we ignore `-r ""`, trailing `-r` or other
-     * silly things the user might (inadvertedly) be doing.
-     */
-    module && require(resolveModule(module, { basedir: cwd }));
+    if (module) {
+      /* This check ensures we ignore `-r ""`, trailing `-r`, or
+       * other silly things the user might (inadvertedly) be doing. */
+      require(resolveModule(module, { basedir: cwd }));
+    }
 });
 
 opts._.forEach(function (arg) {
--- a/bin/tape
+++ b/bin/tape
@@ -13,10 +13,11 @@ var opts = parseOpts(process.argv.slice(2), {
 
 var cwd = process.cwd();
 
-/* If only one require is specified, the value of `opts.require`
- * will be a string. This is why we concatenate.
- */
-;[].concat(opts.require).forEach(function(module) {
+if (typeof opts.require === 'string') {
+  opts.require = [opts.require];
+}
+
+opts.require.forEach(function(module) {
     /* The `module &&` ensures we ignore `-r ""`, trailing `-r` or other
      * silly things the user might (inadvertedly) be doing.
      */
--- a/bin/tape
+++ b/bin/tape
@@ -14,7 +14,7 @@ var opts = parseOpts(process.argv.slice(2), {
 var cwd = process.cwd();
 
 if (typeof opts.require === 'string') {
-  opts.require = [opts.require];
+    opts.require = [opts.require];
 }
 
 opts.require.forEach(function(module) {
--- a/bin/tape
+++ b/bin/tape
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+require('source-map-support/register');
 
 var path = require('path');
 var glob = require('glob');
--- a/bin/tape
+++ b/bin/tape
@@ -20,7 +20,7 @@ if (typeof opts.require === 'string') {
 opts.require.forEach(function(module) {
     if (module) {
       /* This check ensures we ignore `-r ""`, trailing `-r`, or
-       * other silly things the user might (inadvertedly) be doing. */
+       * other silly things the user might (inadvertently) be doing. */
       require(resolveModule(module, { basedir: cwd }));
     }
 });
--- a/bin/tape
+++ b/bin/tape
@@ -6,9 +6,9 @@ var parseOpts = require('minimist');
 var glob = require('glob');
 
 var opts = parseOpts(process.argv.slice(2), {
-        alias: { 'r': 'require' },
+        alias: { r: 'require' },
         string: 'require',
-        default: {'r': [] }
+        default: { r: [] }
     });
 
 var cwd = process.cwd();
--- a/package.json
+++ b/package.json
@@ -15,7 +15,9 @@
     "glob": "~5.0.3",
     "has": "~1.0.1",
     "inherits": "~2.0.1",
+    "minimist": "1.2.0",
     "object-inspect": "~1.0.0",
+    "resolve": "1.1.6",
     "resumer": "~0.0.0",
     "string.prototype.trim": "^1.1.1",
     "through": "~2.3.4"
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "inherits": "~2.0.1",
     "object-inspect": "~1.0.0",
     "resumer": "~0.0.0",
-    "source-map-support": "~0.4.0",
     "string.prototype.trim": "^1.1.1",
     "through": "~2.3.4"
   },
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     "glob": "~5.0.3",
     "has": "~1.0.1",
     "inherits": "~2.0.1",
-    "minimist": "1.2.0",
+    "minimist": "~1.2.0",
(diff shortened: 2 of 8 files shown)


Raynos:
I don't understand what did does.

Tape doesn't do anything with source maps. Why add it here ?

mstade:
My apologies, the description wasn't particularly clear. If the code being run is linked to source maps, then any errors thrown will have a call stack that reflects the original code. Without it, it's much more difficult to understand where errors originate. Does that help?

Raynos:
Can you show me what a workflow looks like ?

Would it be something like

```sh
make build # compile my XJS -> real js
node build/test.js
# :( get bad stacktrace
```

I think it might make more sense for some-one to write `source-map-node` binary that runs any node program but pre-loads source maps support.

This way it's opt-in for people who want source map support and we don't have source map hooks enabled globally for all users of tape.

mstade:
That's pretty much the workflow, yeah.

> I think it might make more sense for some-one to write source-map-node binary that runs any node program but pre-loads source maps support.

Hmm, yes, maybe. The problem with more binaries of course is that they pile up. I want to run with source map support, but I *also* want to run a code coverage collector. But I think you're on to something though. Modules like `source-map-support` that have a kind of cross cutting module like `register` can just be loaded like tests would be. In the example above, it becomes:

```bash
make build # compile my XJS -> real js
tape node_modules/source-map-support/register build/test.js
# :) get good stacktrace
```

Except, for whatever reason I can't seem to figure out right now, the above doesn't work. Would it be meaningful to add a `-r,-require` flag to tape, that would then enable this kind of workflow, yet still be opt-in? Something like:

```bash
tape -r source-map-support/register build/test.js
```

I did a quick hack to test this, and it solves my issue of course. Speculating further, this kind of extension point would potentially enable things like running multiple pre-requisites:

```bash
tape -r source-map-support/register -r collect-coverage build/test.js
```

If one wants to parameterize these modules, or do something more esoteric, you can implement it in a local module:

```bash
tape -r ./my-esoteric-prerequisites build/test.js
```

Happy to change this PR, or make a new one, if you think this is within the purview of tape. Like you, I don't much like the added responsibility this PR puts on tape, which I think is brilliant in its limited scope and simplicity. Adding a `-r,--require` flag would make tape a little more complex, but I think the added value is potentially much greater than the cost of such a feature. What do you think?

mstade:
Playing devil's advocate, I suppose you could also argue that you could do stuff like this:

```bash
```

Raynos:
I'm +1 for a `-r` flag. 

Although demonstrating that this is powerful when it comes to for example, combining istanbul & tape together will help get it landed.

If your going to add an options parser to implement `-r` then please use `minimist`.

mstade:
Right on, thanks for the discussion @Raynos – much obliged! I'll carve out some time to implement `-r` in the next few days, ideally before Christmas. I'll just repurpose this PR for it, but if you feel strongly otherwise feel free to close and I'll just open a new one.

mstade:
@Raynos ok I've removed the source map stuff and implemented `-r` and `--require` instead. I also added a couple of tests for it, but I'm not sure of their quality. I've also not changed the documentation any, which is probably something that should also be done.

Any thoughts?

mstade:
I updated the title and description of this PR to reflect the new changes. It makes the conversation look a bit weird, but whatever. :o)

mstade:
Some thoughts:

### Order is not preserved

Consider the following:

```bash
$ tape -r first second -r third
```

Order is preserved for flags and globs separately, so the order of execution in the above example is `first, third, second`. Would it be more intuitive and meaningful to preserve the order entirely? Could this help with things like #126 where the node process never ends because of lingering handles? I.e. if order was preserved, then setup and teardown could be extracted into modules of their own, and common logic shared:

```bash
$ tape -r setup-db test/*.test.js -r teardown-db
```

Sure enough, this can be done *without* `-r` as well, by having setup and teardown files:

```bash
$ tape test/setup-db.js test/*.test.js test/teardown-db.js
```

But this makes it more difficult to share such modules, for instance on a team that does a lot of similar work. I'm not sure which way to lean on this.

### Parameterizing modules

Parameterizing modules can be done using environment variables:

```bash
$ BABEL_ENV=production tape -r babel-register test/*.js
```

Is there a need for parameterization of modules, that can't be solved with environment variables? If so, should the recommended approach be to implement a local module and load that? E.g.:

```bash
$ cat ./my-esoteric-needs
var mod = require('some-module')
mod.installGlobablly({ some: 'parameters' })

$ tape -r ./my-esoteric-needs test/*.js
```

Raynos:
> Order is not preserved

```
$ tape -r first second -r third
```

I expect `first, third, second`. Like first its going to run all the pre requires and then run all the glob patterns.

Working as expected & intended. If you want to feel warm and fuzzy then run `tape -r first -r third second`

Raynos:
> Parameterizing modules

You are stop on; use environment variables or write a local helper.

mstade:
Cool, thanks for the comments @Raynos. I think I agree with you re: the order of execution; it makes sense to run all requires first, regardless of what order they are specified in on the command line.

I'll see about updating the docs as well before I consider my work here done.

package.json
```diff
@@ -15,7 +15,9 @@
     "glob": "~5.0.3",
     "has": "~1.0.1",
     "inherits": "~2.0.1",
+    "minimist": "1.2.0",
```
Raynos: Please use `~`

bin/tape
```diff
@@ -1,12 +1,32 @@
...
+        string: 'require',
+        default: {'r': [] }
+    });
+
+var cwd = process.cwd();
+
+/* If only one require is specified, the value of `opts.require`
+ * will be a string. This is why we concatenate.
+ */
+;[].concat(opts.require).forEach(function(module) {
```
Raynos: nitpick: would be more readable as

```js
if (typeof opts.require === 'string') {
  opts.require = [opts.require];
}
```

Raynos:
This looks good to me.

We need to add a section to the README about this new feature, i can do that after merging if you prefer.

I think we should wait 24 hours before merging to let other reviewers like @substack take a look.

mstade:
Thanks for the review @Raynos, I've incorporated your feedback and added some docs to the readme as well. Feel free to review and suggest any changes, but I think that about covers it – what's your take?

bin/tape
```diff
@@ -1,12 +1,33 @@
...
+
+if (typeof opts.require === 'string') {
+  opts.require = [opts.require];
+}
+
+opts.require.forEach(function(module) {
+    /* The `module &&` ensures we ignore `-r ""`, trailing `-r` or other
+     * silly things the user might (inadvertedly) be doing.
+     */
+    module && require(resolveModule(module, { basedir: cwd }));
```
ljharb: i'd prefer `if (module) { require(resolveModule(module, { basedir: cwd })); }`, rather than abusing boolean operators for control flow :-/

bin/tape
```diff
@@ -1,12 +1,33 @@
...
+
+var cwd = process.cwd();
+
+if (typeof opts.require === 'string') {
+  opts.require = [opts.require];
+}
+
+opts.require.forEach(function(module) {
+    /* The `module &&` ensures we ignore `-r ""`, trailing `-r` or other
+     * silly things the user might (inadvertedly) be doing.
```
ljharb: inadvertently

bin/tape
mstade: Potato/tomato in my book, but happy to change it. Will push an update in a minute.

bin/tape
mstade: Oops! Thanks. :o)

bin/tape
```diff
@@ -1,12 +1,34 @@
...
 
-var path = require('path');
+var resolveModule = require('resolve').sync;
+var resolvePath = require('path').resolve;
+var parseOpts = require('minimist');
 var glob = require('glob');
 
-process.argv.slice(2).forEach(function (arg) {
+var opts = parseOpts(process.argv.slice(2), {
+        alias: { 'r': 'require' },
```
ljharb: any reason these "r"s are quoted?

bin/tape
```diff
@@ -1,12 +1,34 @@
...
+        string: 'require',
+        default: {'r': [] }
+    });
+
+var cwd = process.cwd();
+
+if (typeof opts.require === 'string') {
```
ljharb: you could replace this if block with `var requires = [].concat(opts.require)`, fwiw (that would also avoid mutating `opts`)

bin/tape
mstade: Force of habit, really. I'll change it.

ljharb:
Overall this LGTM! It'd be great to rebase this down to a smaller number of meaningful atomic commits

bin/tape
mstade: I did something similar [previously][1], but @Raynos suggested the change to the `typeof` block instead. I don't really care much either way to be honest.

[1]: https://github.com/substack/tape/pull/224#discussion-diff-48197171R19

mstade:
Sure, if you decide to merge feel free to rebase any which way you'd like! :o)

mstade:
Thanks for the review btw, @ljharb!

bin/tape
ljharb: i guess either way is fine

ljharb:
Ideally we'd merge through the web interface which means you'd need to do any rebasing.

mstade:
With all due respect – I'm cool with contributing a feature and making sure the code is up to your standards so the patch can be landed, but the bike shedding is strong with this one. Regardless, I'll take @Raynos's cue and await any further feedback from @substack (or others.)

ljharb:
Your contribution is appreciated, but what you call "bikeshedding" involves consistency with a project's style and commit history, and is *at the least* as important as new features. Certainly if you're unwilling to rebase the commits, and we'd still like the feature, we can either click merge and deal with the log clutter, or manually squash the commits ourselves, but to dismiss maintainer concerns as "bikeshedding" doesn't seem very helpful :-/

Awaiting further feedback seems best in any case.

mstade:
You're right @ljharb – my apologies. I didn't mean to offend anyone, nor dismiss feedback. I very much appreciate the solid feedback provided.

Raynos:
This looks solid. I think the README is fine.

As mentioned I'll wait ~24 hours before merging but if there's no strong objections I'll merge tomorrow.

ljharb:
:+1: i'll definitely be using this feature

mstade:
I've [rebased and pushed a new version of this branch](https://github.com/substack/tape/compare/master...mstade:require-flag). I didn't want to mess with this branch, for reasons. Let me know how you'd like to proceed if you'd like to merge this. It's equally trivial to make the same changes to *this* branch, as it is to open up a new PR just for merging – your call.

Thanks for the reviews and feedback!

Raynos:
Published 4.3.0

mstade:
Much obliged! :+1:

PR #232, opened 2015-12-29
Add return values to assertion methods

taylor1791:
I am using [jsverify](https://github.com/jsverify/jsverify) for property testing. `jsverify` terminates early if it returns false. I noticed that there was no way to see if the test succeeds or fails in `tape` for me to return `false`. I am not trying to force ideas from other libraries into `tape`, however returning nothing (`undefined`) doesn't seem to provide any benefit. This will allow people to know if the test succeeds for similar needs.

I know that this does not test all the cases, but I could not find a good way to test when it should fail. If you have an idea on how to improve the tests, I can fix them.

Here is what merging this would allow others to do. There are probably other cases when this would provide benefit, but I don't have an exhaustive list.

```
test('magic', function (t) {
  jsc.assert(jsc.forall("number", "number", function(a, b) {
    // I need to return true if the property holds, and false if it is violated. jsverify will then stop
    // generating data. If a specific number pass (default 100), then the property is assumed to "hold".
    // I have no way to know if `t.equal(a, add(a, b))` succeeds or fails to return the proper value for
    // jsverify.
    return t.equal(a, add(a, b));
  }));
  t.end();
});

function add(a, b) { return a + b; }
```

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -264,10 +264,12 @@ Test.prototype._assert = function assert (ok, opts) {
             actual : self._plan - pendingAsserts
         });
     }
+
+    return ok;
 };
 
 Test.prototype.fail = function (msg, extra) {
-    this._assert(false, {
+    return this._assert(false, {
         message : msg,
         operator : 'fail',
         extra : extra
@@ -275,7 +277,7 @@ Test.prototype.fail = function (msg, extra) {
 };
 
 Test.prototype.pass = function (msg, extra) {
-    this._assert(true, {
+    return this._assert(true, {
         message : msg,
         operator : 'pass',
         extra : extra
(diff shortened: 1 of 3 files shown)


ljharb:
in your example, I'm kind of confused why you'd need to assert on forall, when each iteration already contains an assertion. what value does that final assert add?

taylor1791:
`jsc.forall` return a function (think deferred computation) and does not actually run tests. `jsc.assert` runs the computation created by `jsc.forall` multiple times with random data. If you just use `jsc.forall`, no iterations occur. I believe that `assert` has a different definition that what we are accustomed to.

taylor1791:
Closing since it appears the original authors don't want this

ljharb:
There’s now an onFailure event (in the next release); would that suffice?

taylor1791:
It has been a while since I looked at this (almost a year now), but I don't think so. `onFailure` runs after the test fails. In the example above, we need to make the test fail.

I don't you to try to make it work because 1) I no longer need it and 2) no one else has complained about it. Focus your energy where you can make more impact.

Issue #234, opened 2016-01-08
Asynchronous calls to `test` fail to run


KoryNunn:
Test case here: https://github.com/KoryNunn/tape/blob/async-test-setup/test/async.js

It seems that if `test()` is called outside the initial tick, it wont run.

jtlapp:
Yes, support for the `only` constraint requires this. `tape` queues up all of the tests in the current tick and executes them in a subsequent tick. It has to have seen all of the tests before it can execute any tests in order to know if an `only` test appeared among them. If you define a test in a later tick, it will follow the `tape` evaluation of the tests. It may still get executed, depending on timing and intervening tasks, but it may not, and it may not respect `only` either.

KoryNunn:
Yeah I came to that conclusion eventually too, and now we use https://gist.github.com/KoryNunn/a7ddbfb069bbbb2a41b4 as a solution to the problem we had.

PR #235, opened 2016-01-08
Async test setup

KoryNunn:
As per https://github.com/substack/tape/issues/234

Attempting to call additional tests after some delay causes tape to hang. 

This branch resolves by the addition of the `endWaitTime` option.

By default, there is no wait time, and tape acts as usual.

Setting `endWaitTime` to a value eg `50` causes tape to wait 50ms between t `'end'` events before emitting `'done'`. This allows tests that get set up asynchronously after other tests have completed, to run.

`endWaitTime` is a suggestion, there are probably better names for it.

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -63,16 +63,27 @@ Results.prototype.createStream = function (opts) {
         output.queue('TAP version 13\n');
         self._stream.pipe(output);
     }
-    
+
     nextTick(function next() {
         var t;
         while (t = getNextTest(self)) {
             t.run();
-            if (!t.ended) return t.once('end', function(){ nextTick(next); });
+
+            function end(){
+                if(opts.completeWait){
+                    opts.completeWait(function(){
+                        nextTick(next);
+                    });
+                }else{
+                    nextTick(next);
+                }
+            }
+
+            if (!t.ended) return t.once('end', end);
         }
         self.emit('done');
     });
-    
+
     return output;
 };
 
@@ -93,7 +104,7 @@ Results.prototype._watch = function (t) {
     t.once('prerun', function () {
         write('# ' + t.name + '\n');
     });
-    
+
     t.on('result', function (res) {
         if (typeof res === 'string') {
             write('# ' + res + '\n');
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
sound reasonable to me; implementation looks good

substack:
This would make tests very timing sensitive, which sounds like a fragile feature. What's so bad about moving async operations inside the test:

``` js
test('whatever', function (t) {
  asyncSetup(function () {
     // ...
  })
})
```

or moving async setup to a pre-test:

``` js
test('setup', function (t) {
  asyncSetup(function () { t.end() })
})

test('whatever', function (t) {
  // ...
})
```

KoryNunn:
I'm trying to use tape for integration testing, where each set of tests need to run in series, with a db flush in between.

The issue is, because that flush takes some time, tape decides that the tests must be finished.

I'm not a huge fan of this "just wait a bit" solution, and I'm also investigating other solutions.

I'll give the pre-test implementation a go, my only reservation would be having a test that only serves to prevent tape from finishing.

KoryNunn:
@substack I gave the setup test implementation a go, but it doesn't work, because the results only attempt to complete after a t.end, so no tests are run after the setup test begins.

This PR (https://github.com/substack/tape/pull/236) does work for that case, and also solves the test added in the current PR.

Raynos:
I think this is a bad idea.

There are better approaches.

@KoryNunn Please look at https://www.npmjs.com/package/tape-cluster

KoryNunn:
@raynos I agree, please see #236 and let me know what you think.

nathanmarks:
FWIW, this is how I am handling async test setup:

I have the following helper function that runs my DB migrations on an sqlite DB.

You could also run seeds here, or any other async setup.

```
import Test from 'blue-tape';
import Bookshelf from 'lib/db';

export function TestWithDB (name, callback, testFn = Test) {
  return testFn(name, t => {
    return Bookshelf.knex.migrate.latest().then(() => callback(t));
  });
}

TestWithDB.only = function (name, callback) {
  return TestWithDB(name, callback, Test.only);
};
```

PR #236, opened 2016-01-11
Resolves issue where tests would not attempt to complete

KoryNunn:
This is a solution to the same problem that https://github.com/substack/tape/pull/235 addresses.

I reckon this is a better solution, but I'll leave them both open for review for now.

The new test fails in the current version of tape, and passes with the changes proposed.


KoryNunn:
I don't see this being addressed, so here is a workaround I'm using: https://gist.github.com/KoryNunn/a7ddbfb069bbbb2a41b4

ljharb:
This needs rebasing, but it's still worth evaluating I think?

ljharb:
@KoryNunn can you check the "allow edits" checkbox on the right hand side of the PR?

KoryNunn:
Done.

ljharb:
@KoryNunn this has been rebased, but tests appear to be failing for me - `tap.createConsumer` doesn't seem to be available on the version of `tap` we use.

KoryNunn:
Looks like the changes break other tests, and after 3 years I'm not motivated enough to do anything about it.

Doesn't look like it has much requirement since no one else has been involved in this PR.

Someone else can have a go if they want.

KoryNunn:
I had a look at what the issue was and only the skip test fails. It looks like the outer tap ends before the tape test gets a chance to end it's output stream. I am not sure how to get around that.

PR #238, opened 2016-01-19
Test on Node.js v0.10.x, v0.12.x, v4.x, and v5.x

ntwb:


.travis.yml
```diff
@@ -2,5 +2,5 @@ language: node_js
 node_js:
   - "0.10"
   - "0.12"
-  - "iojs"
```
ljharb: there's no need to remove iojs, support for it isn't dropped.

ntwb:
Added back `iojs`

Raynos:
lgtm.

ntwb:
:+1:

Issue #242, opened 2016-01-23
No "at:" line reported on errors when running through babel-node or babel-tape-runner


carpie:
Given a hypothetical test `foo.js`:

```
var test = require('tape');

test('foo', function myTest(t) {
    t.plan(1);
    t.equal(0, 1);
});

```

If I execute with node, everything is fine:

```
$ node foo.js
TAP version 13
# foo
not ok 1 should be equal
  ---
    operator: equal
    expected: 1
    actual:   0
    at: Test.myTest (/home/carpie/tmp/foo.js:5:7)
  ...

1..1
# tests 1
# pass  0
# fail  1

```

However, if I execute under babel-node or babel-tape-runner, the `at:` indicator is lost making the
test that failed difficult to find:

```
TAP version 13
# foo
not ok 1 should be equal
  ---
    operator: equal
    expected: 1
    actual:   0
  ...

1..1
# tests 1
# pass  0
# fail  1
```

I looked into to this and found it was because the babel stack traces did not have full paths in them, so the current _failure line_ detection did not find them.  I will send a pull request with a proposed alternate method of _failure line_ detection that makes this work.

ljharb:
The babel issue seems to have been fixed in https://github.com/babel/babel/pull/3608 and released in v6.14.0.

PR #243, opened 2016-01-23
Alternate method of failure line detection (fix for #242)

carpie:
This is a proposed fix for losing "at:" reporting when running tests through babel-node and babel-tape-runner (and possibly other tools that modify code).

The previous detection method searched the error stack for frames that listed files using absolute paths.  It then weeded out ones that were part of the tape module directory.  This works fine when running under node because the unmodified code reports full paths for source files.  Under babel-node and babel-tape-runner, the source files are reported as single file names (e.g. no path).  The detection does not work when this happens.

The new method checks each error stack entry to see if the frame belongs to 'test.js' (which is always where the assertions come from).  When it finds the first frame that is not 'test.js', it reports that as the failure line ("at:").  Admittedly, it would still not report "at:" entries of the end user chooses 'test.js' as their file name, but that seems a reasonable restriction to me for more reliable failure line reporting.

When I made this change, it unfortunately (fortunately?) caused some of the tape unit tests to start reporting "at:" lines.  Though a bit challenging :) due to specific error line reporting, I updated those in this pull request too.

With this change, running the example from issue #242, I now correctly get:

```
$ babel-node foo.js
TAP version 13
# foo
not ok 1 should be equal
  ---
    operator: equal
    expected: 1
    actual:   0
    at: Test.myTest (foo.js:5:7)
  ...

1..1
# tests 1
# pass  0
# fail  1
```

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -211,39 +211,18 @@ Test.prototype._assert = function assert (ok, opts) {
     if (!ok) {
         var e = new Error('exception');
         var err = (e.stack || '').split('\n');
-        var dir = path.dirname(__dirname) + '/';
-        
         for (var i = 0; i < err.length; i++) {
-            var m = /^[^\s]*\s*\bat\s+(.+)/.exec(err[i]);
-            if (!m) {
-                continue;
-            }
-            
-            var s = m[1].split(/\s+/);
-            var filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[1]);
-            if (!filem) {
-                filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[2]);
-                
-                if (!filem) {
-                    filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
-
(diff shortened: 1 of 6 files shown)


ksmithbaylor:
This is awesome! Anything I can do to help get this merged?

bySabi:
Why is not merged yet?? 
any problem with this PR @substack @carpie @ljharb

I have 'at' problem too using Babel.

I want help too

carpie:
I agree that if upstream projects reported full paths, that would resolve the stated issue (except for tape's own unit tests which the original method ignores since they share the same path.)

That said, I would be perfectly happy with full stack traces in the reporting instead of having to guess which code in a trace is user's code :)

ljharb:
At the least, can issues be filed on the relevant upstream projects, and linked here?

carpie:
I did some checking... It seems this has already been reported against babel: https://phabricator.babeljs.io/T6839

ljharb:
The babel issue seems to have been fixed in https://github.com/babel/babel/pull/3608 and released in v6.14.0.

PR #244, opened 2016-01-30
Travis: Get rid of sudo

kuraga:


ljharb:
Thanks!

ljharb:
Turns out we need this, because the sudoless travis-ci VMs can't build C extensions for iojs 3/node 4 and higher.

PR #261, opened 2016-02-17
add message defaults to .ok() and .notOk()

paulcpederson:
Just noticed in a reporter that there are really nice test defaults for methods like `.equals`, but `.ok()` and `.notOk()` just say `(unnamed assert)`.

This adds the following default messages:

- `.ok()` - "should be truthy"
- `.notOk()` - "should be falsey"

Also updates the tests where they are expecting the old default value.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -302,7 +302,7 @@ Test.prototype.ok
 = Test.prototype.assert
 = function (value, msg, extra) {
     this._assert(value, {
-        message : msg,
+        message : defined(msg, 'should be truthy'),
         operator : 'ok',
         expected : true,
         actual : value,
@@ -315,7 +315,7 @@ Test.prototype.notOk
 = Test.prototype.notok
 = function (value, msg, extra) {
     this._assert(!value, {
-        message : msg,
+        message : defined(msg, 'should be falsy'),
         operator : 'notOk',
         expected : false,
         actual : value,
--- a/test/default-messages.js
+++ b/test/default-messages.js
@@ -1,0 +1,37 @@
+var tap = require('tap');
+var spawn = require('child_process').spawn;
+var trim = require('string.prototype.trim');
+
+tap.test('default messages', function (t) {
+    t.plan(1);
+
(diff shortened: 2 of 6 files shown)


lib/test.js
```diff
@@ -312,7 +312,7 @@ Test.prototype.notOk
 = Test.prototype.notok
 = function (value, msg, extra) {
     this._assert(!value, {
-        message : msg,
+        message : defined(msg, 'should be falsey'),
```
ljharb: nit: `falsy` shouldn't have an `e` imo.

ljharb: also i don't see any tests covering the falsy case

paulcpederson:
@ljharb ok, updated to `falsy`. None of the tests threw an error when I changed the notOk error message. Does that mean there are currently no tests for `notOk`?

ljharb:
@paulcpederson that's the conclusion i would also come to. would you mind adding some?

fwiw I really like this change - it doesn't change the API, just makes two of the assertions more useful by default.

paulcpederson:
cool. I *think* I can figure out how to do that.

paulcpederson:
@ljharb ok I added a new test that just tests all of the defaults. Sorry that took me a bit to figure out the test structure. Testing a testing framework is very meta.

ljharb:
Awesome, this LGTM as a semver-minor change.

Any other collaborators object? If not, I'll merge this in a day or two.

paulcpederson:
@ljharb bump.

Have any other collaborators taken a look?

ljharb:
Thanks for the bump - I'll merge this today or tomorrow.

ljharb:
@paulcpederson actually would you mind rebasing this on the latest master?

paulcpederson:
Not at all.

ljharb:
Thanks, ping me when it's updated and I'll merge it!

paulcpederson:
@ljharb ok, should be rebased now from `master` thanks for your patience, things got a little crazy today!

Issue #68, opened 2014-03-14
Better error stack traces reporting


Raynos:
@chrisdickinson mentioned that tape doesn't print stack traces properly.

Reporting error cases for failing tests in a clean fashion is important for debugging & testing.

@chrisdickinson should provide more feedback and I can write a PR for some (failing) tests around a more descriptive error output and update the renderer

scottcorgan:
@domenic just published `3.0.0` for tap-spec. Includes some prova-like error reporting.

tschaub:
What are people doing when debugging test failures in a headed browser?  The output below is just from a single failing test.  I can't imagine where I'd start with a failure buried in 100s of passing tests.

![image](https://cloud.githubusercontent.com/assets/41094/7484289/0180eff2-f344-11e4-9629-2639abc6345d.png)

(The failing assertion is in test.js line 6, and there is an inline sourcemap in the test.js bundle.)

Raynos:
@tschaub I try to reproduce it in node and then use that workflow.

if its a browser specific bug then i use `test.only` after I find it.

gritzko:
My findings so far: row.file points to a random location inside tape, which is not helpful.
I send failed test stack traces to the console, so I can click on them.
![tape-dom](https://cloud.githubusercontent.com/assets/81581/8984306/65769f4c-36d9-11e5-9033-f563b5c82c0b.png)

tcurdt:
tape-spec is nice - but (at least for me) does not provide stacktrace or file and line number either.

ljharb:
This seems closed in #265 / #330.

PR #265, opened 2016-02-25
show the entire stack when there's an error

devTristan:
related to #68 

I just made it print the entire stack trace.

Sorry about the removal of the trailing spaces, I can't figure out how to get my editor to not do that when I save.

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -156,8 +156,10 @@ function encodeResult (res, count) {
     if (res.at) {
         output += inner + 'at: ' + res.at + '\n';
     }
-    if (res.operator === 'error' && res.actual && res.actual.stack) {
-        var lines = String(res.actual.stack).split('\n');
+    var error = (res.actual instanceof Error ? res.actual : res.error)
+    var stack = typeof error === 'object' ? error.stack : null
+    if (stack) {
+        var lines = String(stack).split('\n');
         output += inner + 'stack: |-\n';
         for (var i = 0; i < lines.length; i++) {
             output += inner + '  ' + lines[i] + '\n';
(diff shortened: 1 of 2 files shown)


lib/results.js
```diff
@@ -93,7 +93,7 @@ Results.prototype._watch = function (t) {
     t.once('prerun', function () {
         write('# ' + t.name + '\n');
     });
-    
+
```
ljharb: please revert all these whitespace changes; this is part of the project's style

lib/results.js
```diff
@@ -127,23 +127,28 @@ Results.prototype.close = function () {
 function encodeResult (res, count) {
     var output = '';
     output += (res.ok ? 'ok ' : 'not ok ') + count;
-    output += res.name ? ' ' + res.name.toString().replace(/\s+/g, ' ') : '';
-    
+
+    if (typeof res.name === 'object' && typeof res.name.stack === 'string') {
+      output += ' ' + res.name.stack;
+    } else if (res.name) {
+      output += ' ' + res.name.toString().replace(/\s+/g, ' ');
```
ljharb: `String(foo)` is more reliable than `foo.toString()`, please use that instead

devTristan:
:+1: fixed the whitespace and switched to `String(res.name)`

devTristan:
@ljharb [ gentle ping ]

ljharb:
@devTristan sorry about the delay! Would you mind rebasing on latest master? I think this will also need tests ensuring that the stack trace behavior appears when expected, and not when not.

devTristan:
@ljharb alright, I've done the rebase and added a test. Unfortunately this change breaks other tests which do not expect to see a stack trace when an assertion fails. I don't have time to fix them right now, but hopefully I can get to it soon.

OliverJAsh:
This can be closed now with #330 merged?

PR #266, opened 2016-02-29
Skipped test blocks should output a “SKIP” message

ljharb:
Using the `skip` option to skip a test block currently outputs nothing - whereas `t.skip` outputs as a comment prefixed with "SKIP ". This adds the same behavior to when skipping a test block.


ljharb:
If there's no objections (@raynos @substack) I'll plan to merge this sometime tomorrow (3/1)

PR #268, opened 2016-03-03
Ensure that non-functions passed to `throws` fail the test, just like `assert`

### LINKED ISSUE #267: `t.throws(undefined)` passes, and should not
What am I doing wrong here? `t.throws()` doesn't seem to be working (not failing the test when a function is not throwing any exceptions).

Here's a sample test, that passes successfully (instead of failing as expected):

```javascript
'use strict'

var test = require('tape')

function emptyFunction () {
  // nothing happens here
}

test('This test should fail', function (t) {
  t.plan(1)
  t.throws(emptyFunction())
})
```

I'm on Node 5.0, `tape` version `4.4.0`.

ljharb:
Fixes #267

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -458,7 +458,7 @@ Test.prototype['throws'] = function (fn, expected, msg, extra) {
         caught.error = caught.error.constructor;
     }
 
-    this._assert(passed, {
+    this._assert(typeof fn === 'function' && passed, {
         message : defined(msg, 'should throw'),
         operator : 'throws',
         actual : caught && caught.error,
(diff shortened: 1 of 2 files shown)


ljharb:
Two choices:
 - This will be a breaking change, and will necessitate a major version bump
 - this is a bug fix, since `assert` has always failed in these situations, and will be only a patch bump. (in this scenario, I will absolutely backport it to all of the major lines, and as many of the minors as i have patience for)

Thoughts?

substack:
This seems like a patch update if it's fixing unexpected behavior.

ljharb:
:+1: thanks, i'll merge and release this as a series of patch updates soon then

PR #275, opened 2016-03-24
Add test for deep loose equal

brendanlong:
I was experimenting with reducing tape's dependencies and realized that `_.isEqual` doesn't offer a loose equal, but the tests were still passing. This test checks that loose equal works right.


ljharb:
Thanks!

PR #276, opened 2016-03-24
Reduce the number of dependencies

brendanlong:
I was reading about overuse of tiny packages in Node.js and thought it would be fun to try to reduce the dependencies in tape. Pulling in Underscore.js and adding a util.js for `trim()` and `defined()` removes 10 dependencies, reduces the size of `node_modules` by 500 K, and reduces the install time (`npm install --prod`) by half. The "string.prototype.trim" package is the worst offender, pulling in 9 dependencies when MDN gives a one-liner to implement it. The other packages would be more complicated to remove (without 

`npm ls --prod` before:

    tape@4.5.1 /Users/blong/workspace/tape
    ├── deep-equal@1.0.1
    ├── defined@1.0.0
    ├── function-bind@1.1.0
    ├─┬ glob@7.0.3
    │ ├─┬ inflight@1.0.4
    │ │ └── wrappy@1.0.1
    │ ├─┬ minimatch@3.0.0
    │ │ └─┬ brace-expansion@1.1.3
    │ │   ├── balanced-match@0.3.0
    │ │   └── concat-map@0.0.1
    │ ├── once@1.3.3
    │ └── path-is-absolute@1.0.0
    ├── has@1.0.1
    ├── inherits@2.0.1
    ├── minimist@1.2.0
    ├── object-inspect@1.1.0
    ├── resolve@1.1.7
    ├── resumer@0.0.0
    ├─┬ string.prototype.trim@1.1.2
    │ ├─┬ define-properties@1.1.2
    │ │ ├── foreach@2.0.5
    │ │ └── object-keys@1.0.9
    │ └─┬ es-abstract@1.5.0
    │   ├─┬ es-to-primitive@1.1.1
    │   │ ├── is-date-object@1.0.1
    │   │ └── is-symbol@1.0.1
    │   ├── is-callable@1.1.3
    │   └── is-regex@1.0.3
    └── through@2.3.8

After:

    tape@4.5.1 /Users/blong/workspace/tape
    ├── deep-equal@1.0.1
    ├─┬ glob@7.0.3
    │ ├─┬ inflight@1.0.4
    │ │ └── wrappy@1.0.1
    │ ├─┬ minimatch@3.0.0
    │ │ └─┬ brace-expansion@1.1.3
    │ │   ├── balanced-match@0.3.0
    │ │   └── concat-map@0.0.1
    │ ├── once@1.3.3
    │ └── path-is-absolute@1.0.0
    ├── inherits@2.0.1
    ├── minimist@1.2.0
    ├── object-inspect@1.1.0
    ├── resolve@1.1.7
    ├── resumer@0.0.0
    ├── through@2.3.8
    └── underscore@1.8.3

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,9 @@
-var defined = require('defined');
 var createDefaultStream = require('./lib/default_stream');
 var Test = require('./lib/test');
 var createResult = require('./lib/results');
+var defined = require('./lib/util').defined;
 var through = require('through');
+var _ = require('underscore');
 
 var canEmitExit = typeof process !== 'undefined' && process
     && typeof process.on === 'function' && process.browser !== true
--- a/lib/results.js
+++ b/lib/results.js
@@ -3,9 +3,9 @@ var inherits = require('inherits');
 var through = require('through');
 var resumer = require('resumer');
 var inspect = require('object-inspect');
-var bind = require('function-bind');
-var has = require('has');
-var regexpTest = bind.call(Function.call, RegExp.prototype.test);
+var _ = require('underscore');
+
+var regexpTest = _.bind(Function.call, RegExp.prototype.test);
 var yamlIndicators = /\:|\-|\?/;
 var nextTick = typeof setImmediate !== 'undefined'
     ? setImmediate
@@ -39,7 +39,7 @@ Results.prototype.createStream = function (opts) {
                     name: t.name,
                     id: id
                 };
-                if (has(extra, 'parent')) {
+                if (_.has(extra, 'parent')) {
                     row.parent = extra.parent;
                 }
                 output.queue(row);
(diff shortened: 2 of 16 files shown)


brendanlong:
I also opened pull request #275  because I initially thought that I could replace `deep-equal` with `_.isEqual` and the unit tests didn't catch the mistake.

index.js
```diff
@@ -1,8 +1,9 @@
-var defined = require('defined');
 var createDefaultStream = require('./lib/default_stream');
 var Test = require('./lib/test');
 var createResult = require('./lib/results');
+var defined = require('./lib/util').defined;
 var through = require('through');
+var _ = require('underscore');
```
ljharb: i absolutely do not want to pull in underscore - even if this was an improvement, we'd want lodash over underscore a thousand times over.

ljharb:
Optimizing for the number of dependencies adds no value - one possible argument would be the *file size* of the dependency tree, but more deps is better. In the case of underscore/lodash, I'd want the separate lodash modules, *not* the entire grab bag of lodash.

brendanlong:
Ah thanks, Lodash does look much nicer.

That's reasonable. I think tape is small enough that looking through dependencies was less useful than I was hoping. I would still have a preference for a bunch of lodash dependencies just for the single upstream and the fact that lodash doesn't depend on anything outside of itself, but it wouldn't make a huge difference.

Issue #223, opened 2015-12-10
"test exited without ending"


gritzko:
Hi!

Probabilistically, I get the following:

```
not ok 95 test exited without ending
  ---
    operator: fail
  ...
not ok 96 test exited without ending
  ---
    operator: fail
  ...

(more of the same)

    operator: fail
  ...
not ok 112 test exited without ending
  ---
    operator: fail
  ...
not ok 113 test exited without ending
  ---
    operator: fail
  ...

1..113
# tests 113
# pass  94
# fail  19
```

19 out of 20 runs are OK.
I am mostly annoyed by the fact I can not localize/attribute the error.

joshrivers:
@gritzko Is this on Node, or in the browser?

Raynos:
This means your test did not call `assert.end()`.

Probably a race condition in your own code.

kgryte:
I am seeing a similar issue. I have tests split across multiple files. When I run the tests using `Istanbul`

``` bash
$  istanbul cover tape -- ./test/*.js 
```

some runs all complete (no errors) and others return an `operator: fail` error due to `test exited without ending`. In my case, the failed test resides in the last test file.

The tests are all synchronous, so no async race conditions. And all tests explicitly call `t.end()`.

Marak:
Are you certain your tests are idempotent and that the state of some tests
aren't affecting the state of other tests?

What happens if you run the tests individually? Do they work every time? If
so, the issue could be state related.

gritzko:
@joshrivers node, AFAIR
@kgryte yes, looks like my case, there was explicit t.end() and the case was synchronous
I switched to tap, so I can not provide any more feedback.

kgryte:
@Marak If I run each test file individually, the tests __always__ pass. No tests share state.

Marak:
@kgryte -

Can you describe the nature of the tests? If they are all synchronous tests I'm assuming they are performing no file-system or network based i/o? Or are they using synchronous file-system methods?

What I'm trying to ascertain is if you are accessing any shared resources across the tests which could be causing conflicting state.

kgryte:
The tests are of the variety:

``` javascript
test( 'awesome test', function ( t ) {
  t.ok( 5 === 5, 'numbers are equal' );
  t.end();
});
```

In the last file which occasionally fails, I added additional (synchronous) dummy test cases to see what happens. The result is that either all tests pass or all fail due to the `test exited without ending` error.

Marak:
[...]

PR #278, opened 2016-03-28
Loop over all files after resolving glob patterns

kl0tl:
Closes https://github.com/substack/tape/issues/223.

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -3,7 +3,9 @@
 var resolveModule = require('resolve').sync;
 var resolvePath = require('path').resolve;
 var parseOpts = require('minimist');
-var glob = require('glob');
+var glob = require('globby');
+
+var nextTick = require('../lib/next_tick');
 
 var opts = parseOpts(process.argv.slice(2), {
         alias: { r: 'require' },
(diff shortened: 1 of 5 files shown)


Kl0tl:
That’s weird, [test/require.js](https://github.com/substack/tape/blob/master/test/require.js) is failing on node 0.10. It works with `t.plan(1)` instead of `t.end()` in [test/require/a.js](https://github.com/substack/tape/blob/master/test/require/a.js) though, but I’m not really sure why. Would it be enough to fix the failing test @ljharb ?

ljharb:
I think we should assume the test is correct.

ljharb:
Could you add a test that fails *without* your change?

kgryte:
@Kl0tl Any updates on this?

jcama:
Hi! @Kl0tl I'm working on a similiar test framework project. Wont to help me out? If so drop an email at jonnycasamillo@gmail.com. For others as well. Feel free to send an email!

r0mflip:
Seemed to be fixed by https://github.com/substack/tape/pull/374

Issue #222, opened 2015-12-10
Test order is affected by a presence of a nested test


ArtemGovorov:
The following code:

``` JavaScript
var test = require('tape');

test('first', function (t) {

  t.plan(2);

  setTimeout(function () {
    t.ok(1, 'first test');
  }, 200);

  t.test('second', function (t) {
    t.ok(1, 'second test');
    t.end();
  });
});

test('third', function (t) {
  setTimeout(function () {
    t.ok(1, 'third test');
    t.end();
  }, 100);
});
```

outputs:

```
ok 1 first test
ok 2 second test
ok 3 third test
```

and it's expected. But if I change it to its logical equivalent without `plan`:

``` JavaScript
var test = require('tape');

test('first', function (t) {

  setTimeout(function () {
    t.ok(1, 'first test');
    t.end();
  }, 200);

  t.test('second', function (t) {
    t.ok(1, 'second test');
    t.end();
  });
});

test('third', function (t) {
  setTimeout(function () {
    t.ok(1, 'third test');
    t.end();
  }, 100);
});
```

the the test order is broken:

```
ok 1 second test
ok 2 third test
ok 3 first test
```

While changing the order of the first and the second test is _somewhat_ expected, ending the first test prematurely, and letting the third test execute before it, looks like a bug. Especially considering that commenting `t.test('second'...` test out restores the expected order, and make the third test wait until the first one ends.

davegomez:
Is anyone checking at this? @substack @ljharb

GuidoS:
I think this may be tied to this other open issue https://github.com/substack/tape/issues/122

Raynos:
This is not an ordering problem; this is a reporting problem.

`t.end()` will wait for all child tests to finish.

I agree this is confusing semantic difference between `t.end()` and `t.plan()`.

I recommend adding `console.log()` to convince yourself of the actual order of execution and to not worry about which order tape reports TAP statements to stdout.

ArtemGovorov:
@Raynos From what I see, it is definitely an ordering issue, not a reporting one.
Adding `console.log` (with timestamps just in case), produces the same result:

``` JavaScript
var test = require('tape');

test('first', function (t) {

  setTimeout(function () {
    console.log('first test', +new Date());
    t.ok(1);
    t.end();
  }, 200);

  t.test('second', function (t) {
    console.log('second test', +new Date());
    t.ok(1);
    t.end();
  });
});

test('third', function (t) {
  setTimeout(function () {
[...]

PR #279, opened 2016-03-28
Document unexpected `t.end()` behavior

kl0tl:
See https://github.com/substack/tape/issues/222 for motivation.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -408,6 +408,74 @@ With [npm](https://npmjs.org) do:
 npm install tape --save-dev
 ```
 
+# troubleshooting
+
+Sometimes `t.end()` doesn’t preserve the expected output ordering.
+
+For instance the following:
+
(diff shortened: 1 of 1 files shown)


abhisekp:
Please merge this! 
Required!!

abhisekp:
@Kl0tl I believe `t.plan` should be used for any async tests. Please correct me If i'm wrong.

abhisekp:
Basically, `t.plan` must be used whenever there is an async in the test and in all other cases which doesn't involve async, `t.end` can be used.

yoshuawuyts:
Yup, yeah pretty much - for consistency I never use `t.end()`

ljharb:
@kl0tl can you check the "allow edits" checkbox on the right hand side of the PR?

ljharb (approved):

PR #283, opened 2016-04-09
Update readme.markdown

supercrabtree:


ljharb:
This module isn't a TAP reporter, it's a monkeypatch of tape itself. If it belongs anywhere, it's "other".

PR #293, opened 2016-06-13
[results] increase object-inspect depth for expected/actual

scottbessler:
It may be best for this to be configurable, but for now 5 seems like
a default that is often too low resulting in confusing output that shows
expected/actual to be identical when the difference is >5 deep.

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -141,8 +141,8 @@ function encodeResult (res, count) {
     output += inner + 'operator: ' + res.operator + '\n';
     
     if (has(res, 'expected') || has(res, 'actual')) {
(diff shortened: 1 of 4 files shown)


ljharb:
I think if we're going to make a change like this, we might as well allow it to be configurable.

Personally I think 5 is way way too deep to avoid confusion, let alone 10 - so clearly that number is subjective.

scottbessler:
I hear ya. Made it configurable via opts.objectPrintDepth

ljharb:
This seems reasonable to me, thanks (it'll be a semver-minor).

I'll leave this here for a few more days before merging.

PR #295, opened 2016-07-12
Fix Windows path support in error messages

### LINKED ISSUE #220: No line number provided in Windows on failed tests (and proposed fix)

hgwood:
Supersede #221. Fix #220. Generally makes tape support more kinds of stacktrace formats.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -6,6 +6,12 @@ var EventEmitter = require('events').EventEmitter;
 var has = require('has');
 var trim = require('string.prototype.trim');
 
+// regex taken from https://github.com/errwischt/stacktrace-parser/blob/fb58c13ba19dd9aa12b8480d8e9b938524953921/lib/stacktrace-parser.js#L13
+// and edited to:
+// - have a matching group around everything after the 'at'
+// - exclude stracktraces where the location is internal to Node.js (i.e. is only a filename)
+var stacktraceRegexp = new RegExp('^\\s*at ((?:((?:\\[object object\\])?\\S+(?: \\[as \\S+\\])?) )?\\(?(.*?\\' + path.sep + '.*?):(\\d+)(?::(\\d+))?\\)?\\s*)$');
+
 module.exports = Test;
 
 var nextTick = typeof setImmediate !== 'undefined'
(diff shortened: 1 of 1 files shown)


hgwood:
The stacktrace that is now picked up in some tests is internal to Node and dependent on the version of Node, so the tests break on everything but Node 6. I'm going to tweak the regex to exclude this case.

hgwood:
If fixed by excluding stacktrace elements where the location is only a filename. It must contain at least one path separator.

ljharb:
I think this change will need extensive tests to prevent against current and future breakage.

hgwood:
What test cases would you like to see?

ljharb:
Things that verify important aspects of current stack traces to ensure that a regex-based approach (which is hard to get right) isn't breaking anything. These tests may already exist, of course, but I suspect they don't.

hgwood:
Right. So right in fact, the regexp I'm using here may only work in Node envs. I was focused on that, but tape is supposed to work in browsers too right? [Here](https://github.com/errwischt/stacktrace-parser/blob/master/lib/stacktrace-parser.js#L13) we can different regexp for different envs.

What about using https://www.npmjs.com/package/stacktrace-parser?

Alternatively, one quick and dirty fix I tested for Windows is to use `path.sep` [here](https://github.com/substack/tape/pull/295/files#diff-dd2e881af3311f834199c7b515cff0c7R226) and replace `\/` by `(?:\/|[A-Z]:\\` [here](https://github.com/substack/tape/pull/295/files#diff-dd2e881af3311f834199c7b515cff0c7L229) (and in the 2 other occurrences in the next lines). I did not take that route in the end because it just adds a special case instead of being general. For example, Windows network paths would not work with this. However, I would settle for it if you would be more comfortable with it.

ljharb:
@hgwood Given #316, should this PR be closed, or rebased and re-evaluated?

hgwood:
I will close it as I'm personally satisfied with #316. Anyone willing to go the extra mile to support special paths (like @bsuh) can always open a new PR.

Issue #299, opened 2016-07-17
.only runs all tests with same name


jtlapp:
I have a fix for this already, just wanted to give it an issue # before PRing.

`.only` identifies the test to run by the name of the test. If the developer (accidentally) writes multiple tests with the same name and one of those tests is declared `.only`, all tests having this duplicated name are run.

There are two main approaches to solving this. We can enforce test name uniqueness, or we can generate a unique ID for each test and use that ID. The former would be helpful to the developer, but the latter is simpler, faster, and makes a good setup for a desperately needed feature whereby we can run a single test without having to muck with the code. (I'll post this feature shortly.)

jtlapp:
Can someone make sure my fix doesn't introduce a race condition? A race condition could potentially happen if multiple test modules are loaded asynchronously AND the increment operator is not atomic. We might then have multiple tests with the same ID. See line 48 of `lib/test.js`.

jtlapp:
This SO answer says that ++ is not guaranteed to be atomic, so the next question is, does tape otherwise support asynchronous loading? http://stackoverflow.com/questions/680097/ive-heard-i-isnt-thread-safe-is-i-thread-safe/680114#680114

ljharb:
That's in C++ - JS is single-threaded. You don't have to worry about it.

jtlapp:
Ah yes! Thanks for the reminder. I actually program in Java for Android too, where I have to worry about these things.

I've written a way to pass a command line `-n` to indicate the number of a test to run, with `-n` adding test numbers to the test titles. However, I'm having trouble writing the test suite. I've emulated `test/require.js` for my tests, but I keep getting the following:

`Warning: Possible EventEmitter memory leak detected. 11 close listeners added. Use emitter.setMaxListeners() to increase limit`

Any suggestions for me? Then I'll write the feature request and make its PR. Thanks!

jtlapp:
(My `end` event is being called, but not the `data` event. Code is otherwise directly analogous to `test/require.js` and `test/require/*.js`, as far as I can tell.)

jtlapp:
(Okay, nevermind. My `-n` broke streaming, so gotta look at that.)

PR #300, opened 2016-07-17
.only no longer runs all tests having the same name

jtlapp:
This is a fix for #299, wherein all tests having the same name as the test marked `.only` are run.

### DIFF
--- a/index.js
+++ b/index.js
@@ -138,9 +138,10 @@ function createHarness (conf_) {
     var only = false;
     test.only = function (name) {
         if (only) throw new Error('there can only be one only test');
-        results.only(name);
+        var t = test.apply(null, arguments);
+        results.only(t.number);
         only = true;
-        return test.apply(null, arguments);
+        return t;
     };
     test._exitCode = 0;
     
--- a/lib/results.js
+++ b/lib/results.js
@@ -83,8 +83,8 @@ Results.prototype.push = function (t) {
     self.emit('_push', t);
 };
 
-Results.prototype.only = function (name) {
-    this._only = name;
+Results.prototype.only = function (number) {
+    this._only = number;
 };
 
 Results.prototype._watch = function (t) {
@@ -176,7 +176,7 @@ function getNextTest (results) {
     do {
         var t = results.tests.shift();
         if (!t) continue;
-        if (results._only === t.name) {
+        if (results._only === t.number) {
             return t;
         }
     } while (results.tests.length !== 0)
--- a/lib/test.js
+++ b/lib/test.js
@@ -13,6 +13,7 @@ var nextTick = typeof setImmediate !== 'undefined'
     : process.nextTick
 ;
 var safeSetTimeout = setTimeout;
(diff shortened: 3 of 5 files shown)


ljharb:
I wonder if it might be a better idea to just store a reference to the actual test function that's marked "only", rather than adding unique IDs?

jtlapp:
That would work too, but I would then undo and redo it with numbers for the feature I've just written. I now have a command line argument `-n` for showing test numbers and `nX` for running only test number X. It works when manually tested, but I'm struggling to write the test suite.

ljharb:
I'm skeptical about the utility of that - that immediately makes test ordering tightly coupled to the user's expectations, and adding a test in the middle of the suite would renumber everything.

The numbers appear for TAP output purposes, but I don't think it's a good idea to rely on them beyond referencing a specific test in a specific test run.

jtlapp:
Everything you say is true, but I don't understand the use case where these problems arise. My use case is that I've run the test suite and have one test I want to work on. My options are:

(1) Insert `.only` and hope I remember to remove it before checking the source back in or rerunning the test suite with expectations of testing everything. Any time I edit source, I also risk accidentally inserting or deleting characters, just because the mouse is what it is.
(2) Use the extremely ephemeral `-n` to work this one test case right in the moment without taking risks with source code. Once I've got that one working, I don't care about the number ever again.

jtlapp:
It's also an option. Leave `-n` off and no tests get numbered, and people who prefer to modify old test code to get it working can still do so. The option makes people like me more comfortable using the test framework. I really am loath to be changing code that shouldnt' change.

jtlapp:
I was hoping to have this discussion under its own issue, so I created one #301. I only mentioned the feature here to explain my approach to the present problem.

ljharb:
Let's definitely keep discussing in #301.

As for this one, I think this is a good thing to fix - but I'd prefer not to implement the machinery for #301 until that's the direction decided upon.

Would you be willing to rework this to store the function reference, using object identity to determine which test is the "only" one?

jtlapp:
I agree that if we're not going to be identifying tests by number, then the `.only` feature can be reference-based.

If we went ahead with a reference-based solution now and later decided to implement `-n`, then the code would either implement two different ways to select the test to run -- reference and number -- or we rip out the reference solution and restore the number solution.

Given that I only found this problem in pursuit of the `-n` implementation, I think it's safe to say that there's no urgency to fixing it. People use unique test names.

So I'm thinking that the resolution to this problem can wait on a decision for #301. I just don't see that we should be spending time doing something that we might undo.

jtlapp:
Besides, people are going to know it the moment that they put `.only` on a test that is not uniquely named, because they'll get two test results. They would then change one of the names and be on their way. So this bug isn't much of a problem.

jtlapp:
I'm curious why this pull request is still open if you've already merged the code.

jtlapp:
Oh, this was my original number-based proposal. Closing.

Issue #301, opened 2016-07-18
Command line option -n to run a test by number


jtlapp:
I've written this and having it working but can't seem to get a test suite running. UPDATE: It's now working at the link below. Just waiting for decision on PR.

I absolutely love the simplicity of `tape`, but it asks me to do something that I find anathema: modify code to run a particular test and remember to remove my modifications when I'm done. If I forgot to remove the `.only` constraint, I may check it in. That's easy to fix. The worst scenario is unwittingly changing something in the test while it's in the editor. This happens.

So while I respect that some people are really comfortable with the `.only` feature, I suspect I'm not the only one who is really uncomfortable with it. Fortunately, there's an easy fix.

All we need is a `-n` command line option. If you leave the option off, everything runs as it has. If you include a simple `-n`, all the tests run as usual, except that a test number is prefixed to the title of each test.

To run a particular test, just place `-nN` on the command line, where `N` is the test number shown in the title.

This approach allows `tape` to satisfy die-hards like me who prefer not to make edits that shouldn't be checked in.

jtlapp:
I've written it already, and you can play with it. However, the test suite `test/only-number.js` does not work, and I have yet to figure out why. The `data` events aren't being emitted. Help is appreciated! You'll find it here: https://github.com/jtlapp/tape/tree/only-by-number

jtlapp:
Okay, I found the problem. I load the `tape` module from the module in which the test in installed, in order to communicate the desired test number to it, but when testing the `tape` module itself, there is not `tape` module installed. Hmm....

jtlapp:
The test suite is now working. You can see the proposal at https://github.com/jtlapp/tape/tree/only-by-number. Let me know if you'd like me to do a PR.

jtlapp:
Okay, I think I now have the readme updated and prettied for this proposal.

jtlapp:
For the record, my `-n` solution depends on my fix for issue #299.

jtlapp:
Here is how to modify faucet to support -n. https://github.com/jtlapp/faucet/tree/tape-options-support

jtlapp:
`-n` also facilitates team development. I can now check code into a branch and ask someone, such as somebody remote, to help me debug a particular test number. However, I suppose if we were using different operating systems and multiple test files, there's a chance that the numbers may not correspond. I suspect they usually will, though.

jtlapp:
[...]

PR #302, opened 2016-07-26
-n option and fix for .only on dup test names

jtlapp:
This PR provides the `-n` option described in #301 and the fix for #299 on which it depends.

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -6,9 +6,10 @@ var parseOpts = require('minimist');
 var glob = require('glob');
 
 var opts = parseOpts(process.argv.slice(2), {
-        alias: { r: 'require' },
+        alias: { r: 'require', n: 'number' },
         string: 'require',
-        default: { r: [] }
+        boolean: 'n',
+        default: { r: [], n: false }
     });
 
 var cwd = process.cwd();
@@ -24,6 +25,11 @@ opts.require.forEach(function(module) {
       require(resolveModule(module, { basedir: cwd }));
     }
 });
+
+if (opts.number !== false) {
+    var tape = require(resolveModule("tape", { basedir: cwd }));
+    tape.setNumbering(opts.number);
+}
 
 opts._.forEach(function (arg) {
     glob(arg, function (err, files) {
--- a/bin/tape
+++ b/bin/tape
@@ -27,7 +27,13 @@ opts.require.forEach(function(module) {
 });
 
 if (opts.number !== false) {
-    var tape = require(resolveModule("tape", { basedir: cwd }));
+    var tape;
+    try {
+        tape = require(resolveModule("tape", { basedir: cwd }));
+    }
+    catch(err) {
+        tape = require("../"); // assume testing tape module itself
+    }
     tape.setNumbering(opts.number);
 }
 
--- a/index.js
+++ b/index.js
@@ -138,9 +138,10 @@ function createHarness (conf_) {
     var only = false;
     test.only = function (name) {
         if (only) throw new Error('there can only be one only test');
-        results.only(name);
+        var t = test.apply(null, arguments);
+        results.only(t.number);
         only = true;
-        return test.apply(null, arguments);
+        return t;
     };
     test._exitCode = 0;
     
--- a/index.js
+++ b/index.js
@@ -82,7 +82,7 @@ function createExitHarness (conf) {
             var only = harness._results._only;
             for (var i = 0; i < harness._tests.length; i++) {
                 var t = harness._tests[i];
-                if (only && t.name !== only) continue;
+                if (only && t.number !== only) continue;
                 t._exit();
             }
         }
@@ -107,6 +107,7 @@ function createHarness (conf_) {
     if (conf_.autoclose !== false) {
         results.once('done', function () { results.close() });
     }
+    results.only(onlyTestNumber);
     
     var test = function (name, conf, cb) {
         var t = new Test(name, conf, cb);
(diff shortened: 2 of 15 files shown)


bin/tape
```diff
@@ -25,6 +26,17 @@ opts.require.forEach(function(module) {
     }
 });
 
+if (opts.number !== false) {
+    var tape;
+    try {
+        tape = require(resolveModule("tape", { basedir: cwd }));
+    }
+    catch(err) {
+        tape = require("../"); // assume testing tape module itself
```
ljharb: can you elaborate on this code block? adding production code just for tests seems very subpar.

index.js
```diff
@@ -119,8 +120,11 @@ function createHarness (conf_) {
                 if (!r.ok && typeof r !== 'string') test._exitCode = 1
             });
         })(t);
-        
+    
         results.push(t);
+        if (t.number === onlyTestNumber)
+            results.only(onlyTestNumber);
```
ljharb: please add curly braces around all if blocks, in this case for consistency with lines 107-109.

index.js
```diff
@@ -138,13 +142,23 @@ function createHarness (conf_) {
     var only = false;
     test.only = function (name) {
         if (only) throw new Error('there can only be one only test');
-        results.only(name);
         only = true;
-        return test.apply(null, arguments);
+        var t = test.apply(null, arguments);
+        if (!onlyTestNumber)
+            results.only(t.number);
```
ljharb: please add curly braces around all if blocks, in this case for consistency with lines 107-109.

index.js
```diff
@@ -138,13 +142,23 @@ function createHarness (conf_) {
...
     return test;
 }
+
+var onlyTestNumber = false;
+
+function setNumbering (number) {
+    Test.showTestNumbers();
+    if (typeof number === "number")
+        onlyTestNumber = number;
+}
```
ljharb: please ensure all files have a trailing newline

lib/test.js
```diff
@@ -44,8 +46,11 @@ function Test (name_, opts_, cb_) {
 
     var args = getTestArgs(name_, opts_, cb_);
 
+    this.number = nextTestNumber++;
     this.readable = true;
     this.name = args.name || '(anonymous)';
+    if (numbered)
+        this.name = '['+ this.number +'] '+ this.name;
```
ljharb: please add curly braces around all if blocks

lib/test.js
```diff
@@ -499,5 +504,9 @@ Test.skip = function (name_, _opts, _cb) {
     return Test(args.name, args.opts, args.cb);
 };
 
+Test.showTestNumbers = function() {
```
ljharb: please add a space after "function"

jtlapp:
Oops, forgot to check for consistency with existing style.

Also, my colleague found a bug whereby indicating a test number greater than the number of tests results in all tests running. I'm thinking that I should either run no tests in this case and do one of the following:

- output a message to the console (stdout, I presume)
- output a message to stderr
- output nothing

ljharb:
I think that should error out without running any tests, yes.

What if I run a number that's explicitly skipped? I assume it would still be skipped.

jtlapp:
I agree, it should still be skipped. I'll make some more unit tests. Thank you!

jtlapp:
Another to-do item for me: Make sure the bracketed numbers are compatible with the TAP standard.

jtlapp:
When I've got this done, should I close this PR and make a new one? (I'm new to git/github.)

ljharb:
@jtlapp no, please do not. Rebase and force push and the PR will be updated.

jtlapp:
Okay, I'll try to make sense of that. Out of curiosity, why a trailing newline?

ljharb:
http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

jtlapp:
Thanks. That explains why I have a habit of ending all my text/data files in a newline, but I guess it's been a while since I've catted or munged code via command line tools.

bin/tape
jtlapp: By way of explanation, the `-n` option must inform the tape module of the number of the test to run. The above `resolveModule()` call requires that a module named `tape` have been installed. No module named `tape` has been installed when we run the test suite for `tape` itself. In order to be able to test the `-n` option within the `tape` module itself, it's necessary to load the module in the way that is required by the test suite. You'll find all the tests loading the `tape` module this way as well (not just mine). Without this line, the `tape` module would not be able to include tests of the `-n` option.

ljharb: It seems like we could resolve that by adding a devDep of `"tape": "./"`?

jtlapp: I don't know. That's trick NPM stuff. That might allow you to replace the mysterious `var tape = require('../');` line found in each test file.

jtlapp:
Presently, if you use `-n` for a test that is being skipped, you get the same result as doing this using your current code base:

`test.only("skip or only?", {skip: true}, (t) => { ... }`

Which is this:

    # SKIP lexically invalid domain name
    
    1..0
    # tests 0
    # pass  0
    
    # ok

The numbering shows that a test is being skipped -- numbers are skipped. I like that. It's informative.

I'm getting my `-n` implementation right before looking for how to hook it in.

bin/tape
ljharb: That line isn't mysterious - that's how every npm package tests itself.

jtlapp:
I'm closing this. I found an easy way to extend the `tap` module to do the job. Thanks for your help!

PR #303, opened 2016-08-01
reference-based .only fix for #299

jtlapp:
.only now identifies tests by reference instead of by test name, fixing #299

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -22,7 +22,7 @@ function Results () {
     this.pass = 0;
     this._stream = through();
     this.tests = [];
-    this._only = 0;
+    this._only = null;
 }
 
 Results.prototype.createStream = function (opts) {
--- a/lib/results.js
+++ b/lib/results.js
@@ -22,6 +22,7 @@ function Results () {
     this.pass = 0;
     this._stream = through();
     this.tests = [];
+    this._only = 0;
 }
 
 Results.prototype.createStream = function (opts) {
@@ -83,8 +84,8 @@ Results.prototype.push = function (t) {
     self.emit('_push', t);
 };
 
-Results.prototype.only = function (name) {
-    this._only = name;
+Results.prototype.only = function (t) {
+    this._only = t;
 };
 
 Results.prototype._watch = function (t) {
@@ -176,7 +177,7 @@ function getNextTest (results) {
     do {
         var t = results.tests.shift();
         if (!t) continue;
-        if (results._only === t.name) {
+        if (results._only === t) {
             return t;
         }
     } while (results.tests.length !== 0)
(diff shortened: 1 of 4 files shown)


lib/results.js
```diff
@@ -22,6 +22,7 @@ function Results () {
     this.pass = 0;
     this._stream = through();
     this.tests = [];
+    this._only = 0;
```
ljharb: this can probably be initialized to `null`, or `function () {}`, rather than zero?

jtlapp: Yes, my preference is to initialize to null. I was trying to be consistent with the rest of the code, as you had asked. ;-)

ljharb: The other things that init to 0 tho end up assigned to numbers, right? ie, the convention is to initialize such that the type of the property doesn't change?

jtlapp: Okay, just checked through them. Maybe I was confusing those with another project. I'll fix. To minimize the chance of me screwing up, it would help for you to point me to the right way to update a PR. I screwed up a PR on a project once so that thereafter every single PR indicated failed merges despite merges really being clean. I'll spend an hour researching before making an attempt. I've gotten git-shy.

ljharb: np - basically you just rebase your local branch as you like - without deleting the branch or pushing any changes to the remote - and then, assuming "origin" is the remote that points to your fork, `git push origin ref-based-only -u -f` - without the "-f" it will refuse to push it.

jtlapp: Sorry, what did you want to accomplish with the rebase? I'm getting `Current branch ref-based-only is up to date.` My `master` branch has edits, so I'm working off of a `source` branch that's a copy of your `master`. That's the message I get on `git rebase source`.

ljharb: You'd need to do a fetch off the `source` remote, and then `git rebase source/master`

jtlapp:
It appears to have updated properly. Now to add notes to my already-huge set so notes on git.

jtlapp:
I can't `git rebase upstream/master` from this merge, even after doing a `git fetch upstream`. I get this, indicating that it doesn't see the changes you just merged in:

```
--- a/index.js
+++ b/index.js
@@ -138,9 +138,10 @@ function createHarness (conf_) {
     var only = false;
     test.only = function (name) {
-        results.only(name);
+        var t = test.apply(null, arguments);
+        results.only(t.number);
         only = true;
-        return test.apply(null, arguments);
+        return t;
     };
     test._exitCode = 0;
```

jtlapp:
I attempted another approach that also didn't work:

- I created a new branch and pulled from upstream/master.
- I merged in my hook changes into this branch.
- I attempted to `git rebase upstream/master` but once again got unreasonable merge conflicts.

jtlapp:
Okay I figured it out. I had to walk the rebase forward through time, even if it meant temporarily adding back in code that has since been deleted.

PR #305, opened 2016-08-02
Added optimal settings for travis

amilajack:


.travis.yml
```diff
@@ -1,8 +1,18 @@
...
-  - "0.10"
-  - "0.12"
-  - "iojs"
-  - "4"
-  - "5"
+  - 6
+  - 5
+  - 4
+  - 0.12
+  - 0.10
```
ljharb: these MUST be surrounded in quotes, or else travis will interpret this as "0.1".

.travis.yml
```diff
@@ -1,8 +1,18 @@
...
+  - iojs
+
+cache:
+  directories:
+    - node_modules
```
ljharb: this is actually not optimal, because anything in the dep tree that has a semver-included update released will then mean that "what's tested" is not the same as "what's installed on a clean install".

node_modules should only ever be cached on travis for shrinkwrapped and/or top-level apps.

.travis.yml
```diff
@@ -1,8 +1,18 @@
...
+
+install:
+  - npm install -g npm@latest
```
ljharb: similarly, we *want* to test the stock npm version, and not blindly update to latest. `tape` needs to work when installed with npm 2 as well as npm 3.

ljharb:
Thanks for the contribution! However, we already have an optimal travis config.

Issue #304, opened 2016-08-02
Flow control hooks for alternatives to tape command


jtlapp:
I proposed a `-n` command line argument to number tests and to run only a single test by number, where a "test" refers to a named group of assertions. @ljharb and I discussed this proposal in #301 but decided that the proper solution would change output too much to merit incorporating it into `tape` proper. So @ljharb then suggested that I look into ways to add hooks to `tape` to support a third party command that would do the job.

I've investigated this and found it quite feasible. I already have a solution figured out. I'm therefore creating this "issue" to discuss whether and how to proceed.

I have patched a solution into a fork of `tape`, rather than into a third party module, just to get things working. I discuss this patch starting in [this comment](https://github.com/substack/tape/issues/301#issuecomment-236731945).

I now propose creating a version of `tape` that has my `onPreTest()` hook for `-n`, with tests, but which behaves as `tape` currently does.

I would also like a hook for causing the test run to stop on the first failed test. That is, it would stop after completing the first run of a named test whose assertions do not all pass. I plan to support that in my third party command with an option such as `-s`. This would probably be a hook on test results, to give the handler flexibility to react based on the results.

There is one more hook I would like. This one is potentially more controversial. I'm finding in my test suites that occasionally I forget to include cleanup code at the end of a test. This affects the next test, and I spend some time figuring out the careless mistake. I would prefer to just hook in a cleanup routine that always runs after each test, no matter what. Peace of mind. Note that this proposal is different from the business of running a setup routine prior to each test. I opted for `tape` in part because I believe setup should always be explicit. (The cleanup hook could be called `onPostTest()`, complementing `onPreTest()`.)

If I proceed with this, I do have a question about how to test my third party command for all proper `tape` command behavior. I would prefer to run the `tape` test suite against my command, but the suite is hard-coded to run `bin/tape`. Any suggestions there? Thanks!

jtlapp:
I now have this working for a separate command that hooks into the `tape` module, but `tape` has a modelling decision that introduces more complexity than I would have liked.

`tape` semantically distinguishes skipping a test via `.skip` and skipping tests via the `skip` option. If you use `.skip`, the test has no effect on the output. If you use the `skip` option, a line is output saying that the test was skipped. If there were no distinction, I could just pass an argument to my hook indicating whether a test was scheduled in any fashion. But because the distinction exists, I have to either tell the hook about the distinction (arguments `included` and `skipped`) or have special processing in `tape` proper. Both work, but neither pleases me.

[...]

PR #308, opened 2016-08-05
Pre- and post-processing hooks, plus abort/bail-out

jtlapp:
These are the simplest modifications I can find for allowing me to hook in the features I need for [`tapeo`](https://www.npmjs.com/package/tapeo). See #304


jtlapp:
I'm closing this. I found an easy way to extend the `tap` module to do the job. Thanks for your help!

Issue #312, opened 2016-08-26
devDependencies out-of-date. intentional? 


nelsonic:
Hi Fellow Tapers, 

we are just busy finishing writing a "_why you should use tape_" post and noticed that the `devDependecies` are out-of-date https://david-dm.org/substack/tape?type=dev  [![devDependencies Status](https://david-dm.org/substack/tape/dev-status.svg)](https://david-dm.org/substack/tape?type=dev)

![tape-out-dependencies-out-of-date](https://cloud.githubusercontent.com/assets/194400/18007163/ceb9d0cc-6b9a-11e6-9745-9b692708c399.png)

Is this intentional or, can we submit a PR updating them?

yoshuawuyts:
oh yeah pretty sure it's not intentional - PR welcome

nelsonic:
When we update to the latest versions of `tap` and `tap-parser` we get quite a few failing tests:
![tape-tests-fail-when-devdependencies-updated](https://cloud.githubusercontent.com/assets/194400/18007480/2a1b7cbc-6b9c-11e6-98d2-7e79b20d890e.png)

So there might be some re-work required for a PR.... worth it?

yoshuawuyts:
While not crucial it'd be appreciated I reckon

ljharb:
The failing tests are indeed the reason they aren't updated. It would be quite appreciated to get a minimal PR to update them.

nelsonic:
Hi @yoshuawuyts & @ljharb (_and the rest of the Tape Team!_),
have made the necessary update to the _first_ test that needs to be updated to work with the _latest_ version of Tap. please see: https://github.com/nelsonic/tape/commit/7db3bdcc5ca5b1f7be6ccf112a4438ed47be65c3

followed the example found in this PR: https://github.com/substack/tape/pull/171/files which was merged by @Raynos 

When the test is run we get:

<img width="473" alt="tape-array-test-passes-after-update" src="https://cloud.githubusercontent.com/assets/194400/18030921/1caaf572-6cc2-11e6-9b76-201e0b9ed098.png">


Just to _confirm_: am I OK to continue updating the tests as necessary to bring them into line with the latest `tap` which matches the Tap Spec: https://testanything.org/tap-specification.html ? 

Let me know and I will finish off the update and submit a PR today. (_thanks!_)

ljharb:
Those changes seem reasonable, thanks! Other than trying to keep diffs as minimal as possible, I'd love to review that PR.

[...]

PR #314, opened 2016-08-28
Update to latest version of tap and tap-parser

nelsonic:
This pull request updates `devDependencies` to the _latest_ version of `tap` & `tap-parser`
( as discussed in: https://github.com/substack/tape/issues/312 )

Had to remove `tap.createConsumer` in favour of `concat-stream` which is being used in: [`test/circular-things.js#L3`](https://github.com/substack/tape/blob/092344b906cd3774ba1812d2db431e77916c9f19/test/circular-things.js#L3)

Hopefully the changes are self-explanatory and not too much `diff`.


test/default-messages.js
```diff
@@ -1,37 +1,34 @@
...
+            '',
+            ''
         ]);
-    });
-
-    var ps = spawn(process.execPath, [ __dirname + '/messages/defaults.js' ]);
-    ps.stdout.pipe(tc);
-});
+    }));
+});
```
ljharb: please make sure all files end in a trailing newline

test/end-as-callback.js
```diff
@@ -64,3 +59,25 @@ function fakeAsyncWrite(name, cb) {
 function fakeAsyncWriteFail(name, cb) {
     cb(new Error("fail"))
 }
+
+function getStackTrace (rows) {
```
ljharb: no space after a function's name

test/end-as-callback.js
```diff
@@ -1,38 +1,33 @@
...
-            "fail  1"
+            'TAP version 13',
+            '# do a task and write',
+            'ok 1 null',
+            'ok 2 should be equal',
+            '# do a task and write fail',
+            'ok 3 null',
+            'ok 4 should be equal',
+            'not ok 5 Error: fail',
+            getStackTrace(rs), // see: https://git.io/v6hGG
```
ljharb: let's move this URL comment to the function itself - also, it seems like the sort of thing we'll want to reuse. can/should it be put in a separate file and required in?

ljharb:
So far this looks good.

test/end-as-callback.js
nelsonic: So far haven't needed to re-use it. but as soon as it becomes useful in another file I will split it out and `require`. 👍

nelsonic:
@ljharb build passing. ready for feedback. (_thanks!_)

test/end-as-callback.js
```diff
@@ -1,38 +1,32 @@
...
-            { id: 4, ok: true, name: "should be equal" },
-            { id: 5, ok: false, name: "Error: fail" },
-            "tests 5",
-            "pass  4",
-            "fail  1"
-        ])
+        var rs = rows.toString('utf8');
 
+        tt.equal(rs,
+           'TAP version 13\n'
```
ljharb: these might read cleaner if they were arrays with `.join('\n')`?

ljharb: or alternatively, `.split('\n')` on line 10

nelsonic: @ljharb yeah, I saw _both_ approaches in the tests:
[`test/circular-things.js#L11-L27`](https://github.com/substack/tape/blob/092344b906cd3774ba1812d2db431e77916c9f19/test/circular-things.js#L11-L27) vs: [`test/skip.js#L11-L21`](https://github.com/substack/tape/blob/092344b906cd3774ba1812d2db431e77916c9f19/test/skip.js#L11-L21) vs: [`stackTrace.js#L29-L51`](https://github.com/substack/tape/blob/092344b906cd3774ba1812d2db431e77916c9f19/test/stackTrace.js#L29-L51) ... happy to go with _either_, which is _preferable_?

ljharb: Given those three, I think https://github.com/substack/tape/blob/092344b906cd3774ba1812d2db431e77916c9f19/test/skip.js#L11-L21 is the cleanest

nelsonic: Ok, cool I will loop through all the files this evening and change them to match the `.join('\n')` sample in `test/skip.js`

nelsonic:
@ljharb tests updated to match the format in `test/skip.js` as discussed in https://github.com/substack/tape/pull/314/files#r76536364

test/default-messages.js
```diff
@@ -1,37 +1,32 @@
 var tap = require('tap');
 var spawn = require('child_process').spawn;
-var trim = require('string.prototype.trim');
+var concat = require('concat-stream');
 
 tap.test('default messages', function (t) {
     t.plan(1);
 
-    var tc = tap.createConsumer();
+    var ps = spawn(process.execPath, [ __dirname + '/messages/defaults.js' ]);
```
ljharb: let's use `path.join` here - ideally the `/` is never hardcoded, so things work on Windows.

test/exit.js
```diff
@@ -1,38 +1,33 @@
...
+            'ok 4 should be equivalent',
+            'ok 5 should be equivalent',
+            '',
+            '1..5',
+            '# tests 5',
+            '# pass  5',
+            '',
+            '# ok',
+            '',
+            ''
```
ljharb: it seems kind of weird to have all these double-blank-lines at the end - is that expected?

test/default-messages.js
nelsonic: @ljharb, thanks for the tip on cross-platform compatibility. 👍 
do we need to remove _all_ forward slashes? 
or is it enough to go from:
```js
var ps = spawn(process.execPath, [ __dirname + '/messages/defaults.js' ]);
```
To:
```js
var ps = spawn(process.execPath, [ require('path').join( __dirname, 'messages/defaults.js') ]);
```

From @domenic's Gist: https://gist.github.com/domenic/2790533#paths-and-urls I _deduce_ that this should be enough.

test/exit.js
nelsonic: @ljharb _agree_. I thought it was weird too!
However when we _remove_ those two double-black lines in _this_ test we see:
tape-test-exit-remove-last-two-lines-fails

And given that Travis-CI is happy with the the assertion _with_ the double-blank-lines I get _some_ confidence that is how tap output is cross-platform... 😕

test/default-messages.js
nelsonic: _And_ if we change this _one_ instance, should we change _all_ (**Four**) occurrences: 
https://github.com/substack/tape/search?utf8=%E2%9C%93&q=spawn%28process.execPath ? (_just say the word and I'll update them all to be consistently windows-friendly_)

ljharb: That gist uses slashes in `path.relative` (which also won't work, I think), but no, we'd want `path.join(__dirname, 'messages', 'defaults.js')` (you can verify because `path.join('a', 'b\\c')` doesn't result in `a/b/c`, so you can assume that the reverse wouldn't hold in Windows). Yes, it'd be great to update them all.

test/exit.js
ljharb: Can we verify if that's expected tap output behavior, or is that an issue with streams, or something else?

test/default-messages.js
domenic: The reverse does hold in Windows.

```
Domenic@Revan MINGW64 ~/Dropbox/Programming/WIP/custom-elements-copier (master)
$ node
> path.join("a", "b/c")
'a\\b\\c'
```

ljharb: Ah, that's fascinating, and good to know. So unix paths get special treatment that windows paths don't?

domenic: The Windows OS treats / as \ in cases like this, so Node chooses to normalize since doing so has no effect besides making the string consistent. (POSIX does not act symmetrically.)

ljharb: @nelsonic in that case, `path.join( __dirname, 'messages/defaults.js')` is fine throughout. Thanks!

test/exit.js
nelsonic: @ljharb yeah, this is already present in `master` see: [`test/double_end.js#L25`](https://github.com/substack/tape/blob/092344b906cd3774ba1812d2db431e77916c9f19/test/double_end.js#L25) if I update the `test/exit.js` to _match_ this will that be OK? ❓

nelsonic:
@ljharb updates made as discussed in the in-line comments. https://github.com/substack/tape/pull/314#discussion_r76667817 _Thanks_.

nelsonic:
@ljharb please let me know if anything else is required. (_thanks again!_)

test/exit.js
ljharb: Sure, that sounds fine.

nelsonic: Updates made in: https://github.com/substack/tape/pull/314/commits/c4025c181d10a62e972e2d83246cde18b96f5b1b
let me know if OK. 👍

test/default-messages.js
```diff
@@ -1,37 +1,31 @@
...
 var spawn = require('child_process').spawn;
-var trim = require('string.prototype.trim');
+var concat = require('concat-stream');
 
 tap.test('default messages', function (t) {
     t.plan(1);
 
-    var tc = tap.createConsumer();
+    var ps = spawn(process.execPath,
+        [ require('path').join(__dirname, 'messages', 'defaults.js') ]);
```
ljharb: i'd prefer this indentation to be such that if the entire invocation doesn't fit on one line, each argument goes on a line by itself, as does the closing invocation paren.

nelsonic: @ljharb like this:
```js
var ps = spawn(
  process.execPath,
  [ require('path').join(__dirname, 'messages', 'defaults.js') ]
);
```

ljharb: The first example is exactly what I had in mind.

Personally I find "max line length" restrictions silly. If your first example is too long for you, you could do:
```js
var path = require('path');
var ps = spawn(
  process.execPath,
  [
    path.join(__dirname, 'messages', 'defaults.js')
  ]
);
```

(in either case, please move the require to a variable at the top level)

nelsonic: Good point! 
Went with the single line with the `require('path') moved above its only 89 chars which I think is _acceptable_:
```js
var ps = spawn(process.execPath, [path.join(__dirname, 'messages', 'defaults.js')]);
```

ljharb:
Thanks, I think this looks great, and the work is very appreciated!

If you don't mind, it would be great to rebase this down to remove all the "tidy" kind of commits, so that each commit represents a single conceptual change - if that's too much trouble, it's OK, but it'd be nice :-)

After that I'll merge this!

nelsonic:
@ljharb see: https://github.com/substack/tape/pull/318 (_is that OK?_)

ljharb:
Thanks, that'll be fine - although I'd have preferred to merge this as-is rather than open a second PR

PR #316, opened 2016-08-31
Fix for not showing path for error messages on windows

mkls:
This bug has been around for a while now, there are two open PR's (#221 and #295) but no activity recently.
This is basically the solution proposed by @hgwood.
Tried to keep changes to the bare minimum.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -217,7 +217,7 @@ Test.prototype._assert = function assert (ok, opts) {
     if (!ok) {
         var e = new Error('exception');
         var err = (e.stack || '').split('\n');
-        var dir = path.dirname(__dirname) + '/';
+        var dir = path.dirname(__dirname) + path.sep;
         
         for (var i = 0; i < err.length; i++) {
             var m = /^[^\s]*\s*\bat\s+(.+)/.exec(err[i]);
@@ -226,12 +226,12 @@ Test.prototype._assert = function assert (ok, opts) {
             }
             
             var s = m[1].split(/\s+/);
-            var filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[1]);
+            var filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[1]);
             if (!filem) {
-                filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[2]);
+                filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[2]);
                 
                 if (!filem) {
-                    filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
+                    filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
 
                     if (!filem) {
                         continue;
--- a/test/require.js
+++ b/test/require.js
@@ -65,5 +65,5 @@ function tape(args) {
   var proc = require('child_process')
   var bin = __dirname + '/../bin/tape'
 
-  return proc.spawn(bin, args.split(' '), { cwd: __dirname })
+  return proc.spawn('node', [bin].concat(args.split(' ')), { cwd: __dirname })
 }
--- a/test/require/a.js
+++ b/test/require/a.js
@@ -1,8 +1,8 @@
 var tape = require('../..');
 
 tape.test('module-a', function(t) {
+  t.plan(1)
   t.pass('loaded module a')
-  t.end()
 })
 
 global.module_a = true
--- a/test/require/b.js
+++ b/test/require/b.js
@@ -1,8 +1,8 @@
 var tape = require('../..');
 
 tape.test('module-b', function(t) {
+  t.plan(1)
   t.pass('loaded module b')
-  t.end()
 })
 
 global.module_b = true


hgwood:
Third time's the charm!

lib/test.js
```diff
@@ -217,7 +217,7 @@ Test.prototype._assert = function assert (ok, opts) {
     if (!ok) {
         var e = new Error('exception');
         var err = (e.stack || '').split('\n');
-        var dir = path.dirname(__dirname) + '/';
+        var dir = path.dirname(__dirname) + path.sep;
```
ljharb: please use `path.join` here instead of directly referencing `path.sep`

lib/test.js
```diff
@@ -226,12 +226,12 @@ Test.prototype._assert = function assert (ok, opts) {
...
             var s = m[1].split(/\s+/);
-            var filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[1]);
+            var filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[1]);
             if (!filem) {
-                filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[2]);
+                filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[2]);
                 
                 if (!filem) {
-                    filem = /(\/[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
+                    filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
```
ljharb: does `path` expose all the various separators directly, so we don't have to hardcode them?

hgwood: It does but using that would mean more impact on the code (mainly compiling the regexp from a string concatenation). Also the drive letter + colon matching would still have to be hard-coded.

lib/test.js
hgwood: Do you mean `path.join(path.dirname(__dirname), path.sep)` ?

lib/test.js
ljharb: alright, this is probably fine then. it might help to have a comment that unpacks the regex a bit, but nbd.

lib/test.js
ljharb: Ah, I see that `path.join` will always omit the trailing separator. This is fine as-is.

ljharb:
Are there tests that ensure this isn't breaking anything, and tests that are currently failing on windows?

mkls:
As far as I can tell there were no specific tests for this feature, but this modification does not break the current tests. (On windows one test was breaking even before this, I can try to fix that if needed).

If a well tested, multi-platform stack-trace parser is desired, then we should probably include stacktrace-parser as suggested in #295.
I feel like a larger refactor would introduce more possibilities for errors and the current minimal solution is easier to inspect and verify on code review. Basically you only have to decide if the changed regex would still work. I could not come up with cases where it would not.

I will be on vacation from tomorrow, but if you could give me some guidance which way to go about it, I will try to do what it takes next week.

bsuh:
I'd like to see proper errors in all cases. So I modified tape to `console.log` a stack trace as well as `__dirname`.

# node index.js

 | Stack Trace Path | __dirname
------------ | ------------ | -------------
*nix | /Users/bsuh/tape-errors/node_modules/tape/lib/test.js | /Users/bsuh/tape-errors/node_modules/tape/lib
Windows | \\\\vmware-host\Shared Folders\vagrant\tape-errors\node_modules\tape\lib\test.js | \\\\vmware-host\Shared Folders\vagrant\tape-errors\node_modules\tape\lib

# browserify -d index.js | tape-run

 | Stack Trace Path | __dirname
------------ | ------------ | -------------
*nix | node_modules/tape/lib/test.js | /node_modules/tape/lib
Windows | vmware-host/Shared Folders/vagrant/tape-errors/node_modules/tape/lib/test.js | /vmware-host\Shared Folders\vagrant\tape-errors\node_modules\tape\lib

As I see it there are 3 problems

1. proper errors with browserify is broken, because browserify insists on prefixing `__dirname` with `'/'` on all platforms and chrome's stack traces don't contain prefix `'/'` in the filename.
2. tape's regexp for getting the filename from stack trace expects filename to begin with '/' not working with Windows filename.
3. chrome's stack traces contain unix style paths even on Windows, while browserify uses platform dependent paths for `__dirname`.

ljharb:
@bsuh seems like a couple of those need filing upstream on browserify itself - can you file those, and link them here?

bsuh:
@ljharb 

* Removing of`'/'` prefix from browserify `__dirname` seems like a breaking change. I think it's more practical for tape to handle it.
* There is already a pull request to normalize `__dirname`/`__filename` to Unix style on Windows https://github.com/substack/insert-module-globals/pull/63 Not sure the likelyhood of it actually getting merged, since it's been open nearly a year.

hgwood:
Lets be realistic. This is the *third* PR of this subject. The first two were never merged because one the two parties (the contributor or the maintainer) stopped answering requests/questions from the other party, after the discussion got too long (and that's OK).

The goal of this *third* attempt is to introduce the *minimal* change to fix the bug for *most* cases. The goal is not to fix for all cases (see #295 for that, @ljharb).

Merging this PR is safe. The new regex is guaranteed to match anything the old one matched, so regressions are not possible. The worst than can happen is users *that did not see any stack trace before* seeing a partial stack trace. In the meantime, Windows users finally get proper stack traces on errors.

Lets get 80% of the benefits for 20% of the effort. 100% correctness can come later, one step at a time. :)

ljharb:
@mkls would you mind rebasing on latest master so there's no merge commits?

I'll merge this after that.

PR #319, opened 2016-09-07
Add comment formatting

0x1mason:
This PR allows for `util.format` style comment formatting


lib/test.js
```diff
@@ -106,8 +107,11 @@ Test.prototype.test = function (name, opts, cb) {
     });
 };
 
-Test.prototype.comment = function (msg) {
+Test.prototype.comment = function () {
     var that = this;
+    var keys = Object.keys(arguments)
+    var msg = keys.length > 1 ? format.apply(null, arguments) : arguments[keys[0]];
```
ljharb: - just `arguments.length` is sufficient, no need to `Object.keys` (also that makes it require ES5, which this lib does not)
 - It seems like this might be easier to just `format.apply(null, arguments)` in every case?

test/comment.js
```diff
@@ -173,3 +173,57 @@ tap.test('multiline string', function (assert) {
...
+        ].join('\n'));
+    };
+
+    var test = tape.createHarness();
+    test.createStream().pipe(concat(verify));
+    test('formatted multiline string', function (t) {
+        t.comment("tip\n%s\nt%s", "tap", "ape");
+        t.end();
+    });
+});
```
ljharb: please ensure all files always have a trailing newline

0x1mason:
@ljharb Pushed fixes.

lib/test.js
```diff
@@ -106,8 +107,11 @@ Test.prototype.test = function (name, opts, cb) {
     });
 };
 
-Test.prototype.comment = function (msg) {
+Test.prototype.comment = function () {
```
ljharb: i think it'd still be good to leave `msg` here tho, so the function's `.length` is correct :-)

lib/test.js
0x1mason: Calling `format.apply` in all cases changes the behavior, e.g. `test.comment({answer:42})` will print `{answer:42}` whereas the current code prints `[object Object]`. I'm happy to make the change, as I think that's preferable behavior, but it's up to you.

ljharb: Ah, interesting. We definitely want to avoid breaking changes.

Either way, I think perhaps the ES3 equivalent of `format.apply(null, Array.prototype.map.call(arguments, String))` might be the semantic we want here? What's the use case for supporting `format`'s behavior with non-strings?

0x1mason: Use case would be, e.g., `console.log`-style debugging where you want to log object contents. With the change as it stands you can do `t.cmt('%j', obj)`.  Formatting everything, you could do `t.cmt(obj)`.

Doesn't sound like 5 fewer characters is worth the risk of breaking older code tho.

lib/test.js
0x1mason: fixed

lib/test.js
```diff
@@ -108,6 +109,12 @@ Test.prototype.test = function (name, opts, cb) {
 
 Test.prototype.comment = function (msg) {
     var that = this;
+
+    // Previous behavior involved `toString` calls on objects, i.e. emitting `[object Object]`.
```
0x1mason: @ljharb Added a comment to explain the conditional

ljharb: sadly I still don't think this is sufficient - previously, multiple arguments were ignored, and I still might have done `arr.forEach(t.comment)` for example, and relied on the stringification despite passing 3 arguments. I think that the first argument must never be stringified to avoid a breaking change.

0x1mason: Are you suggesting something like this?

```
if args[0] is a string and args.len > 1:
  expand the string
else:
  ignore
```

0x1mason: Also, what about a new function ,eg `format`, `comments`, etc

ljharb: No, I'm suggesting always stringifying the first argument no matter what, and not adding the ability to implicitly log object contents.

ljharb:
@0x1mason could you check the "allow edits" checkbox on the right hand side of the PR?

0x1mason:
@ljharb done

lib/test.js
ljharb: This remains an issue - to restate, I think that we need something like `format.apply(null, Array.prototype.map.call(arguments, String))` to ensure that everything is always stringified.

PR #330, opened 2016-10-17
Show full stack on failure

AnIrishDuck:
I rebased #265 and fixed all the tests. The easiest solution to fixing most tests was to strip out the stack trace parts that will change from machine to machine.

@devTristan @ljharb

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -1,3 +1,4 @@
+var defined = require('defined');
 var EventEmitter = require('events').EventEmitter;
 var inherits = require('inherits');
 var through = require('through');
@@ -157,8 +158,12 @@ function encodeResult (res, count) {
     if (res.at) {
         output += inner + 'at: ' + res.at + '\n';
     }
-    if (res.operator === 'error' && res.actual && res.actual.stack) {
-        var lines = String(res.actual.stack).split('\n');
+
+    var actualStack = res.actual && res.actual.stack;
+    var errorStack = res.error && res.error.stack;
+    var stack = defined(actualStack, errorStack);
+    if (stack) {
+        var lines = String(stack).split('\n');
         output += inner + 'stack: |-\n';
         for (var i = 0; i < lines.length; i++) {
             output += inner + '  ' + lines[i] + '\n';
--- a/test/circular-things.js
+++ b/test/circular-things.js
@@ -2,13 +2,15 @@ var tape = require('../');
 var tap = require('tap');
 var concat = require('concat-stream');
 
+var stripFullStack = require('./common').stripFullStack;
+
 tap.test('circular test', function (assert) {
     var test = tape.createHarness({ exit : false });
     assert.plan(1);
 
     test.createStream().pipe(concat(function (body) {
         assert.equal(
-            body.toString('utf8'),
+            stripFullStack(body.toString('utf8')),
             'TAP version 13\n'
             + '# circular\n'
             + 'not ok 1 should be equal\n'
@@ -18,6 +20,11 @@ tap.test('circular test', function (assert) {
             + '      {}\n'
             + '    actual: |-\n'
             + '      { circular: [Circular] }\n'
+            + '    stack: |-\n'
+            + '      Error: should be equal\n'
+            + '          [... stack stripped ...]\n'
+            + '          at Test.<anonymous> ($TEST/circular-things.js:$LINE:$COL)\n'
+            + '          [... stack stripped ...]\n'
             + '  ...\n'
             + '\n'
             + '1..1\n'
--- a/test/common.js
+++ b/test/common.js
@@ -1,0 +1,56 @@
+var yaml = require('js-yaml');
+
+module.exports.getDiag = function (body) {
+    var yamlStart = body.indexOf('  ---');
+    var yamlEnd = body.indexOf('  ...\n');
+    var diag = body.slice(yamlStart, yamlEnd).split('\n').map(function (line) {
+        return line.slice(2);
+    }).join('\n');
+
+    // The stack trace will vary depending on where the code is run, so just
+    // strip it out.
+    var withStack = yaml.safeLoad(diag);
+    delete withStack.stack;
+    return withStack;
+}
+
+// There are three challenges associated with checking the stack traces included
+// in errors:
+// 1) The base checkout directory of tape might change. Because stack traces
+//    include absolute paths, the stack traces will change depending on the
+//    checkout path. We handle this by replacing the base test directory with a
+//    placeholder $TEST variable.
+// 2) Line positions within the file might change. We handle this by replacing
+//    line and column markers with placeholder $LINE and $COL "variables"
+// 3) Stacks themselves change frequently with refactoring. We've even run into
+//    issues with node library refactorings "breaking" stack traces. Most of
+//    these changes are irrelevant to the tests themselves. To counter this, we
+//    strip out all stack frames that aren't directly under our test directory,
+//    and replace them with placeholders.
+module.exports.stripFullStack = function (output) {
+  var stripped = '          [... stack stripped ...]';
+  var withDuplicates = output.split('\n').map(function (line) {
+      var m = line.match(/[ ]{8}at .*\((.*)\)/);
+
+      var stripChangingData = function (line) {
+          var withoutDirectory = line.replace(__dirname, '$TEST');
+          var withoutLineNumbers = withoutDirectory.replace(/:\d+:\d+/g, ':$LINE:$COL');
+          return withoutLineNumbers;
+      }
+
+      if (m) {
+          if (m[1].slice(0, __dirname.length) === __dirname) {
+              return stripChangingData(line);
+          }
+          return stripped;
+      }
+      return stripChangingData(line);
+  })
+
+  var deduped = withDuplicates.filter(function (line, ix) {
+      var hasPrior = line === stripped && withDuplicates[ix - 1] === stripped;
+      return !hasPrior;
+  });
+
+  return deduped.join('\n');
+}
--- a/test/exit.js
+++ b/test/exit.js
@@ -3,6 +3,8 @@ var path = require('path');
 var spawn = require('child_process').spawn;
 var concat = require('concat-stream');
 
+var stripFullStack = require('./common').stripFullStack;
+
 tap.test('exit ok', function (t) {
     t.plan(2);
 
@@ -38,7 +40,7 @@ tap.test('exit fail', function (t) {
     t.plan(2);
 
     var tc = function (rows) {
-        t.same(rows.toString('utf8'), [
+        t.same(stripFullStack(rows.toString('utf8')), [
             'TAP version 13',
             '# array',
             'ok 1 should be equivalent',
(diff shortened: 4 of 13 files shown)


lib/results.js
```diff
@@ -157,8 +157,10 @@ function encodeResult (res, count) {
     if (res.at) {
         output += inner + 'at: ' + res.at + '\n';
     }
-    if (res.operator === 'error' && res.actual && res.actual.stack) {
-        var lines = String(res.actual.stack).split('\n');
+    var error = (res.actual instanceof Error ? res.actual : res.error)
```
ljharb: `instanceof` will fail on cross-realm Error instances (like from an iframe, or the `vm` module.

lib/results.js
```diff
@@ -157,8 +157,10 @@ function encodeResult (res, count) {
...
+    var stack = typeof error === 'object' ? error.stack : null
```
ljharb: missing semicolon

test/circular-things.js
```diff
@@ -2,13 +2,15 @@ var tape = require('../');
 var tap = require('tap');
 var concat = require('concat-stream');
 
+var stripFullStack = require('./common').stripFullStack
```
ljharb: semicolon

test/common.js
```diff
@@ -0,0 +1,19 @@
...
+
+    // The stack trace will vary depending on where the code is run, so just
+    // strip it out.
+    var withStack = yaml.safeLoad(diag);
+    delete withStack.stack;
+    return withStack;
+}
+
+module.exports.stripFullStack = function (output) {
+    return output.replace(/^\W+at.*:\d+:\d+.*$\W/gm, '');
```
ljharb: Rather than just stripping it, I'd prefer to try to normalize the output across engines - for example, use `try/catch` to get an actual stack trace at runtime, and then using that to figure out how stack traces should work?

test/exit.js
```diff
@@ -3,6 +3,8 @@ var path = require('path');
 var spawn = require('child_process').spawn;
 var concat = require('concat-stream');
 
+var stripFullStack = require('./common').stripFullStack
```
ljharb: ;

test/fail.js
```diff
@@ -3,12 +3,14 @@ var tape = require('../');
 var tap = require('tap');
 var concat = require('concat-stream');
 
+var stripFullStack = require('./common').stripFullStack
```
ljharb: ;

test/common.js
AnIrishDuck: So, here's the first-order problems that checking full stacks with every test would create:

1. The file names will change depending on the absolute directory of the git clone.
2. Any changes to line position (adding or removing things from a test or even potentially to a file that contains code as part of a test traceback) will cause tests to fail.

These are the two biggest problems, and could be easily fixed by stripping the relevant volatile part of the stack. The deeper issue that can't be resolved is the structure of the stack itself. Any refactoring that adds or removes a method to any "failing stack" will now fail the test.

As a result, many trivial refactors would then result in lots of pointless new test failures. It was annoying enough to go through all these tests myself once to fix this issue. I don't want to put that burden on every future author.

Does that make sense? Maybe I'm just not understanding what you're expecting to happen when normalizing output across engines. Is there some deeper property of the stacks we want to test everywhere?

test/circular-things.js
AnIrishDuck: Ugh, sorry. This is what happens when you copy/paste. Will fix all cases.

test/common.js
ljharb: Re 1, we can determine the filename with `__filename` and `process.cwd()` and similar.
Changes to line position are a fair point, and i'd be fine normalizing all of those (ie, all line numbers would be normalized to 0 - so that relative line numbers mattered but not absolute numbers).

I guess I'm mainly concerned (especially because this regex is pretty inscrutable) that there will be bugs in the stack traces, and we won't know.

AnIrishDuck: How about I modify this to something like `verifyAndStripTraces`, which strips all stacks from the generated YAML and checks:

1) That the expected test file shows up somewhere in the stack trace
2) That every line of the stack trace "looks right". Honestly, the only thing I can think of here is matching against some basic regex (or regexes, which might make things more understandable).

For more clarification, here's an example error, with the stacks still fully in the test output:

```
    +++ found                                                                                 
    --- wanted                                                                                
         actual: |-                                                                           
           {}                                                                                 
         stack: |-                                                                            
           Error: should be equivalent                                                        
    +          at Test.assert [as _assert] (/Users/fmurphy/src/tape/lib/test.js:216:54)       
    +          at Test.bound [as _assert] (/Users/fmurphy/src/tape/lib/test.js:65:32)         
    +          at Test.deepEqual.Test.deepEquals.Test.isEquivalent.Test.same (/Users/fmurphy/src/tape/lib/test.js:384:10)
    +          at Test.bound [as deepEqual] (/Users/fmurphy/src/tape/lib/test.js:65:32)       
    +          at Test.<anonymous> (/Users/fmurphy/src/tape/test/undef.js:36:11)              
    +          at Test.bound [as _cb] (/Users/fmurphy/src/tape/lib/test.js:65:32)             
    +          at Test.run (/Users/fmurphy/src/tape/lib/test.js:84:10)                        
    +          at Test.bound [as run] (/Users/fmurphy/src/tape/lib/test.js:65:32)             
    +          at Immediate.next (/Users/fmurphy/src/tape/lib/results.js:71:15)               
    +          at runCallback (timers.js:574:20)                                              
    +          at tryOnImmediate (timers.js:554:5)                                            
    +          at processImmediate [as _immediateCallback] (timers.js:533:5)                  
       ...  
```

If any refactoring adds or removes a function, pulls something into an anonymous function, or moves things into a new file ... then this stack trace will change. Even if we do obvious things, like strip line numbers and character positions, and normalize file paths.

Completely checking stack correctness would require sophisticated code analysis, which seems like overkill. The challenge here is thus figuring out the right tradeoff between "smart" parsing and effort involved.

ljharb: So, I think that it can be simplified by replacing `process.cwd()` with the string "$PWD", for example - and line numbers replaced `:[0-9]+:[0-9]+` with `:#:#` - what other normalization would be needed there?

AnIrishDuck: The problem is, stacks change. All the time. Even running the same test in node version 0.10.0 (which, guessing by the test matrix, you guys want to support) gives a different stack:

```
    +++ found                                                                  
    --- wanted                                                                 
         actual: |-                                                            
           {}                                                                  
         stack: |-                                                             
           Error: should be equivalent                                         
    +          at Test.assert (/Users/fmurphy/src/tape/lib/test.js:216:54)     
    +          at Test.bound [as _assert] (/Users/fmurphy/src/tape/lib/test.js:65:32)
    +          at Test.deepEqual.Test.deepEquals.Test.isEquivalent.Test.same (/Users/fmurphy/src/tape/lib/test.js:384:10)
    +          at Test.bound [as deepEqual] (/Users/fmurphy/src/tape/lib/test.js:65:32)
    +          at Test.<anonymous> (/Users/fmurphy/src/tape/test/undef.js:36:11)
    +          at Test.bound [as _cb] (/Users/fmurphy/src/tape/lib/test.js:65:32)
    +          at Test.run (/Users/fmurphy/src/tape/lib/test.js:84:10)         
    +          at Test.bound [as run] (/Users/fmurphy/src/tape/lib/test.js:65:32)
    +          at Object.next [as _onImmediate] (/Users/fmurphy/src/tape/lib/results.js:71:15)
    +          at processImmediate [as _immediateCallback] (timers.js:309:15)  
       ...  
```

Because, at some point, someone refactored the timers core library in node (and added the `runCallback` and `tryOnImmediate` frames). What normalization method would reduce both stacks to the same value? One has more frames than the other...

AnIrishDuck: We could strip out all frames that don't involve the `test/` directory. So the result would look like this:

```
           Error: should be equivalent                                         
              [... frames omitted ...]
              at Test.<anonymous> (/Users/fmurphy/src/tape/test/undef.js:36:11)
              [... frames omitted ...]
```

Would that be a good compromise?

ljharb: I think that would probably suffice, thanks for investigating!

lib/results.js
```diff
@@ -157,8 +157,10 @@ function encodeResult (res, count) {
     if (res.at) {
         output += inner + 'at: ' + res.at + '\n';
     }
-    if (res.operator === 'error' && res.actual && res.actual.stack) {
-        var lines = String(res.actual.stack).split('\n');
+    var error = (res.actual instanceof Error ? res.actual : res.error);
```
ljharb: I'd prefer not to use `instanceof`, since it's not reliable across realms. (same as https://github.com/substack/tape/pull/330#discussion_r83794224)

lib/results.js
```diff
@@ -157,8 +157,10 @@ function encodeResult (res, count) {
...
+    var stack = typeof error === 'object' ? error.stack : null;
```
ljharb: this won't catch when `error` is `null`

lib/results.js
AnIrishDuck: Right, am still thinking about how to test / fix this. My current plan is to create a new test that throws an error from the node `vm` [module](https://nodejs.org/api/vm.html). Does that sound like a good plan? Any suggestions on actually fixing this issue?

lib/results.js
AnIrishDuck: Will fix.

lib/results.js
ljharb: As a test that's great as long as the test is skipped in non-node environments.

For actually fixing it, it's probably easier to just continue ducktyping the existence of a `stack` property.

AnIrishDuck:
I'm hoping to get to the rest of the comments tomorrow, just was trying to be sure that Travis was happy. Checking errors on node 5 now.

AnIrishDuck:
@ljharb - I think it should be good to go now. I was never able to create a failing test with the `vm` module. For comparison functions i.e. `vm.runInNewContext('t.equal(1,2)')`, the result reporting would happen inside the `vm` realm, and the check would succeed. 

For error-throw checks, I also tried doing:

```js
t.throws(function () {
    vm.runInNewContext('throw new Error(\'CROSS\')');
}, /DOMAIN/);
```

The problem there is that every `_assert` for error testing sets `actual` and `error` to the same exception. Thus, the check was still harmless. Regardless, I think the new code is more foolproof, and doesn't implicate the same realm concerns.

Let me know if modifications are needed.

lib/results.js
```diff
@@ -157,8 +158,12 @@ function encodeResult (res, count) {
     if (res.at) {
         output += inner + 'at: ' + res.at + '\n';
     }
-    if (res.operator === 'error' && res.actual && res.actual.stack) {
-        var lines = String(res.actual.stack).split('\n');
+
+    var actualStack = res.actual && res.actual.stack;
+    var errorStack = res.error && res.error.stack;
+    var stack = defined(actualStack, errorStack);
```
ljharb: why `defined` and not `actualStack || errorStack`? Are we worried about a stack trace that's falsy but not `undefined`?

AnIrishDuck: Nah, it just seemed like the right thing to me. I don't think there will be any practical difference between the two, so I'll switch it if you want.

ljharb: nah, it's not adding a dep so i guess it's fine

AnIrishDuck:
@ljharb - anything needed from me to move this forward? I just checked back and I think I addressed all of your issues?

OliverJAsh:
I'm only seeing one line of the stack trace on the latest version of tape. How do I get this working?

ljharb:
@OliverJAsh a version containing this change has not yet been released.

holgerd77:
Hmm, is it actually possible to make this change optional? We have downgraded our ``tape`` dependency due to this change, since in our setup the stack trace is useless information and this extremely bloats the test output makes debugging much harder.

ljharb:
While I wouldn't say a short stack trace is a part of the contract for tape, if you want to open a new issue, I would support an option that limits the number of lines of a stack trace shown.

holgerd77:
@ljharb Thanks a lot, have done this: https://github.com/substack/tape/issues/397

lib/results.js
```diff
@@ -157,8 +158,12 @@ function encodeResult (res, count) {
     if (res.at) {
         output += inner + 'at: ' + res.at + '\n';
     }
-    if (res.operator === 'error' && res.actual && res.actual.stack) {
-        var lines = String(res.actual.stack).split('\n');
+
+    var actualStack = res.actual && res.actual.stack;
```
nhamer: https://github.com/substack/tape/issues/399

Issue #336, opened 2016-12-10
New aliases "truthy", "falsy"?


perguth:
I declare the need for adding new aliases to `.ok()` and `.notOk()` namely `.truthy()` and `.falsy()` to be self-evident.

What do you think?

ljharb:
`tape`'s API mirrors that of `assert` - which does not have `truthy` or `falsy`. If `assert` adds them, I'll be happy to add them here.

pguth:
We use `truthy` and `falsy` in the README to explain what `ok` and `notOk` means. So I thought it would be straightforward to add the corresponding aliases.

binary-koan:
Personally I find `ok` and `notOk` confusing because I always instinctively interpret them as "the thing I expect" and "the thing I don't expect". Reading back through my tape tests I'm constantly thinking "wait, what's the thing that I consider OK here? - oh hang on, it just means truthy", so to me `truthy` and `falsy` would make things much easier to read.

Totally makes sense that you want to follow the assert API, but tape does already seem to have a lot of aliases for readability (e.g. `equal()` and `is()`)

ljharb:
@binary-koan `assert` already has `equal` - but you're right that `is` and `isNot`, and a few other extra aliases, do exist.

queicherius:
I feel like adding (documented) aliases is not too big of a deal. If you want to swap back to `asset` all you'd have to do is one find/replace since the method signature would match.

ljharb:
Closing per https://github.com/substack/tape/issues/336#issuecomment-266235220.

PR #337, opened 2016-12-12
add `truthy` and `falsy` aliases

perguth:
Discussion started here: https://github.com/substack/tape/issues/336


ljharb:
Thanks! However, closing per https://github.com/substack/tape/issues/336#issuecomment-266235220.

PR #351, opened 2017-03-02
t.throws() broken when awaiting on an async function that throws an error

hudlee:
I can catch await ex1() with a try/catch statement so seems broken.
```
async function ex1() {
  throw new Error('slips by')
}

test('throw inside await async func is not noticed', function(t){
  t.plan(1)
  t.throws( async function(){
    await ex1()
  }, /slips by/)
})
```

This patch makes throws and doesNotThrows notice errors
that get turned into unhandled promise rejections where possible.
I used (more complex) promises to avoid async functions in the code.
The only async functions are in test/throws_harmony.js which you
need to run manually with node --harmony test/throws_harmony.js

There is a version of TAP that works with harmony...

What do you guys think?

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -431,66 +431,97 @@ Test.prototype.notDeepLooseEqual
 };
 
 Test.prototype['throws'] = function (fn, expected, msg, extra) {
-    if (typeof expected === 'string') {
-        msg = expected;
-        expected = undefined;
-    }
-
+    var self = this;
     var caught = undefined;
 
     try {
-        fn();
+      var result = fn();
+      if (result && result instanceof Promise) {
+        result.then(function(){
+          handleCatch(undefined);
+        }).catch(function(err){
+          handleCatch(err);
+        })
+        return;
+      }
     } catch (err) {
-        caught = { error : err };
-        if ((err != null) && (!isEnumerable(err, 'message') || !has(err, 'message'))) {
-            var message = err.message;
-            delete err.message;
-            err.message = message;
-        }
+        return handleCatch(err);
     }
 
-    var passed = caught;
+    function handleCatch (err) {
+      var errorsMatched = false
+      var noErrors = err === undefined
 
-    if (expected instanceof RegExp) {
-        passed = expected.test(caught && caught.error);
-        expected = String(expected);
+      if (typeof fn !== 'function') {
+        var err = new TypeError('fn is not a function')
+        noErrors = true
+      }
+
+      if ((err != null) && (!isEnumerable(err, 'message') || !has(err, 'message'))) {
+          var message = err.message;
+          delete err.message;
+          err.message = message;
+      }
+
+      if (typeof expected === 'string') { //this is a dangerous signature
+          msg = expected; // easy to pass a string when you meant a regex
+          expected = undefined; // then not notice any throw passes
+          errorsMatched = true;
+      }
+      else if (expected instanceof RegExp) {
+          errorsMatched = expected.test(err);
+      }
+      else if (typeof expected === 'function' && err) {
+          err = err.constructor;
+      }
+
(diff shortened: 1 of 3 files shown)


ljharb:
1. `instanceof Promise` is not reliable since a Promise could come from a different realm
1. `Promise` instances are not sufficient, since a promise is anything "thenable"

A simpler approach would be, just use normal Promise API - `.then(t.ok, t.fail)` on the return value of your async function.

hudlee:
Could check for .then `if (result && typeof result.then === 'function')`

Or do a check on module load to see if async function causes a syntax error, and load some code that uses async/await when available (like awaiting on fn() inside throws). I want to make throws return a promise that could be awaited on, so you can completely control the order. Curious what the best way to start using async but still support older versions of node is? Are future versions of tape going to be run through babel?

ex1 is much shorter then ex2 especially when there is other code in the same test:
```
test('ex1', function(t) {
    t.throws(async function(){
      await asyncfunc()
    }, /notok/)
    t.end()
})

test('ex2', function(t) {
    asyncfunc().then(function(){
      t.fail()
      t.end()
    }, function(err){
      t.equal(err.message, 'notok')
      t.end()
    })
})
```

`.then(t.fail, t.ok)` almost makes me happy but the stack seems to missing useful line numbers and
I would need helper functions like `.then(t.fail,whichError)`:
```
function whichError(msg) {
  return function(err){
      if (err.message === msg) {
        t.pass()
      }
      else {
        t.fail()
      }
  }
}
```
However it's accomplished I feel an async/promise aware throw is valuable :)

ljharb:
That would be "rejects". "throws" is a synchronous thing - Promises never throw.

hudlee:
Agreed promises never throw. Seems like try/catch inside an async function is able to catch promise rejections when you use await:
```
async function asyncfunc() {
  throw new Error('exerr')
}

async function main() {
  try {
    var x = await asyncfunc()
    console.log('never executed')
  }
  catch(e) {
    console.log(e)
  }
}
main()
```

You convinced me that the following syntax is good enough for now :)
```
test('happy with this syntax', async function(t){
  try {
    await asyncfunc()
    t.fail('should throw exerr')
  }
  catch (e) {
    t.equal(e.message, 'exerr')
  }
  t.end()
})
```

Adding a .rejects method that only works on await inside async funcs and promise rejections that got returned could be confusing:
```
Test.prototype.rejects = function (fn, expected, msg, extra) {
  var self = this;
  var caught = undefined;
  return new Promise(function(resolve, reject){
      if (typeof expected === 'string') {
          msg = expected;
          expected = undefined;
      }

      var handleCatch = function(){
        var passed = caught;

        if (expected instanceof RegExp) {
            passed = expected.test(caught && caught.error);
            expected = String(expected);
        }

        if (typeof expected === 'function' && caught) {
            passed = caught.error instanceof expected;
            caught.error = caught.error.constructor;
        }

        self._assert(typeof fn === 'function' && passed, {
            message : defined(msg, 'should throw'),
            operator : 'throws',
            actual : caught && caught.error,
            expected : expected,
            error: !passed && caught && caught.error,
            extra : extra
        });
      }

      var possiblePromise = fn()

      if (possiblePromise && typeof possiblePromise.then === 'function') {
        possiblePromise.then(function(){
          handleCatch()
          return resolve()
        }, function(err){
          caught = { error : err };
          if ((err != null) && (!isEnumerable(err, 'message') || !has(err, 'message'))) {
              var message = err.message;
              delete err.message;
              err.message = message;
          }
          handleCatch()
          return resolve()
        })
      }
      else { // non async fn that does not return Promise got passed
        caught = {error:new Error('fn: not async and does not return a promise')}
        handleCatch()
        return resolve()
      }
  })
}
```
Thanks for helping me figure this out.

PR #352, opened 2017-03-06
Import EventEmitter as default from events

caseyWebb:
This was causing some issues when building with rollup for the browser. It looks like the default export has been the `EventEmitter` class since node v4.

I can add support for legacy node versions if you wish.

🍻

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -1,5 +1,5 @@
 var defined = require('defined');
-var EventEmitter = require('events').EventEmitter;
+var EventEmitter = require('events'); EventEmitter = EventEmitter.EventEmitter || EventEmitter;
 var inherits = require('inherits');
 var through = require('through');
 var resumer = require('resumer');
--- a/lib/test.js
+++ b/lib/test.js
@@ -2,7 +2,7 @@ var deepEqual = require('deep-equal');
 var defined = require('defined');
 var path = require('path');
 var inherits = require('inherits');
-var EventEmitter = require('events').EventEmitter;
+var EventEmitter = require('events'); EventEmitter = EventEmitter.EventEmitter || EventEmitter;
 var has = require('has');
 var trim = require('string.prototype.trim');
 var bind = require('function-bind');


ljharb:
Absolutely, support down to node 0.8 must be maintained.

Why is this causing problems with rollup? Browserify handles it just fine.

caseyWebb:
From what I understand... rollup does tree shaking per the ES2015 spec, so it throws errors when the exports and imports are mismatched.

ljharb:
The ES2015 spec doesn't and can't dictate anything about CommonJS modules, so if rollup is having trouble with it, it's a bug with rollup.

lib/results.js
```diff
@@ -129,22 +129,22 @@ function encodeResult (res, count) {
...
     output += '\n';
     if (res.ok) return output;
-    
+
     var outer = '  ';
     var inner = outer + '  ';
     output += outer + '---\n';
     output += inner + 'operator: ' + res.operator + '\n';
-    
+
```
ljharb: please revert all whitespace changes on lines you didn't touch.

lib/results.js
```diff
@@ -14,7 +14,7 @@ var nextTick = typeof setImmediate !== 'undefined'
 ;
 
 module.exports = Results;
-inherits(Results, EventEmitter);
+inherits(Results, EventEmitter.EventEmitter || EventEmitter);
```
ljharb: this `||` should happen up at the top, at the declaration site, rather than at the usage site.

caseyWebb:
I think it very well may be. I'm trying to dig deeper into it.

I believe It's got something to do with that node-builtins polyfill and treeshaking (the polyfill is written in ES2015), but I'm not too sure. I've been fighting this thing for a bit 😅 

I'll clean up that whitespace

caseyWebb:
Ok, I think I know what is happening on rollup's side.

The polyfill attempts to export as both the default, and a named `EventEmitter`.

However, since this is commonjs, it's being transformed to the equivalent of `import EventEmitter from 'events'`, when to access the named member it would have to be `import { EventEmitter } from 'events'. I'm not sure if the commonjs resolver is smart enough to transform something like `var { EventEmitter } = require('events')` to the correct import, but it would break compatibility for older nodes here anyway.

I suppose the best solution may be adding `EventEmitter` as a static property of itself in their polyfill to cover all situations.

caseyWebb:
Moved to the top. It's on one line. I wanted to keep it together without any funky syntax.

lib/results.js
```diff
@@ -1,5 +1,5 @@
 var defined = require('defined');
-var EventEmitter = require('events').EventEmitter;
+var EventEmitter = require('events'); EventEmitter = EventEmitter.EventEmitter || EventEmitter;
```
ljharb: i think two lines is better :-)

however, since `require('events') === require('events').EventEmitter` is true in node 0.12 and later, and we need to maintain compat back to 0.8, it seems like the better choice is to leave this line as it is.

caseyWebb:
I submitted a PR to rollup-plugin-node-builtins — this can probably be closed. Sorry for bothering ya on a Sunday night 😅 

Thanks for being so quick though, I appreciate it!

PR #359, opened 2017-04-16
Update readme.markdown

perguth:
Mention [`flip-tape`](https://github.com/pguth/flip-tape/blob/master/README.md) in the section "other". 🦊

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -129,6 +129,7 @@ By default, uncaught exceptions in your tests will not be intercepted, and will 
 - CoffeeScript support with https://www.npmjs.com/package/coffeetape
 - Promise support with https://www.npmjs.com/package/blue-tape
 - ES6 support with https://www.npmjs.com/package/babel-tape-runner or https://www.npmjs.com/package/buble-tape-runner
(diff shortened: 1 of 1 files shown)


ljharb:
So is this the (deprecated in every other language and testing framework) "should" approach, where it polyfills matchers onto every value?

Does it work with `Object.create(null)` or with objects from other realms? What if I'm doing `for..in` over an object, won't the presence of the matchers suddenly break my code?

pguth:
It only attaches tapes methods to the String prototype (see lines marked with `// eslint-disable-line` https://github.com/pguth/flip-tape/blob/master/index.js).

So in the case of a `for...in` loop over a String object your code would probably break. However I don't think `for...in`-ing over a String is a likely case. Nonetheless I should point that out in the readme.

ljharb:
aha, i misunderstood. so you're just doing `description.matcher(…)` instead of `matcher(…, description)`.

That's definitely much less problematic, so I no longer have a concern about including it in the readme. (I'm still confused why you wouldn't want `matcher(description, …)` to avoid any issues with screwing with `String.prototype`)

PR #370, opened 2017-06-11
updated package.json

EsrefDurna:
updated package glob from 7.1.1 to 7.1.2


ljharb:
Is there a particular bugfix this is needed for? Otherwise, the semver range will auto update to this version when you install.

PR #374, opened 2017-06-24
Fix spurious "test exited without ending"

feross:
Fixes: #223

This PR ensures that all test files are required on the first tick.

To make `tape` happy, all tests need to be queued on the first tick. Tape assumes that tests are done when there's nothing in the queue. 

Right now, the `tape` binary is asynchronously resolving the globs and queuing up the tests as those callbacks are called. This can lead to spurious "test exited without ending" errors.

@Raynos identified this line as the issue a while back https://github.com/substack/tape/issues/223#issuecomment-167920577 Let's finally fix this :)

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -30,6 +30,10 @@ opts._.forEach(function (arg) {
     // Note: `glob.sync` may throw an error and crash the node process.
     var files = glob.sync(arg);
 
+    if (!Array.isArray(files)) {
+      throw new TypeError('unknown error: glob.sync did not return an array or throw. Please report this.');
+    }
+
     files.forEach(function (file) {
         require(resolvePath(cwd, file));
     });
--- a/bin/tape
+++ b/bin/tape
@@ -26,10 +26,12 @@ opts.require.forEach(function(module) {
 });
 
 opts._.forEach(function (arg) {
-    glob(arg, function (err, files) {
-        files.forEach(function (file) {
-            require(resolvePath(cwd, file));
-        });
+    // If glob does not match, `files` will be an empty array.
+    // Note: `glob.sync` may throw an error and crash the node process.
+    var files = glob.sync(arg);
+
+    files.forEach(function (file) {
+        require(resolvePath(cwd, file));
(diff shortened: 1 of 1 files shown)


kgryte:
If the glob fails, the `forEach` line will throw, correct? Is this desired behavior?

bin/tape
```diff
@@ -26,10 +26,13 @@ opts.require.forEach(function(module) {
 });
 
 opts._.forEach(function (arg) {
-    glob(arg, function (err, files) {
-        files.forEach(function (file) {
-            require(resolvePath(cwd, file));
-        });
+    var files;
+    try {
+        files = glob.sync(arg);
```
ljharb: When in the async version would an error be generated by glob, compared to the sync version?

Separately, I appreciate that the current PR mirrors the original version's error handling precisely; but it might be better to let `glob.sync`'s error throw, rather than having an `undefined.forEach` error. I'd consider that change semver-patch, since it's just making the error message more helpful.

kgryte:
For async logic which achieves the same goal (but in a different context), see [here](https://github.com/stdlib-js/stdlib/blob/abc31cf6dfe6c0b0d64728f1aa9d3b30464b7d9e/lib/node_modules/%40stdlib/bench/bin/cli#L133).

bin/tape
feross: > When in the async version would an error be generated by glob, compared to the sync version?

I expect would be generated in the same situations. The docs are unclear about which conditions generate errors. In the case that a glob matches no files, an empty array of files is returned (no error).

I agree with you about `undefined.forEach` being unhelpful. I will update the PR so that `glob.sync` throws. An alternative would be to suppress/ignore any errors (as the original code was doing).

ljharb: To my reading, the original code would still throw the `undefined.forEach` error, it'd just be thrown from the async callback.

ljharb:
I'll cut a release of this later today.

feross:
@ljharb Nice, thanks!

Issue #59, opened 2014-01-20
before/beforeEach/after/afterEach


thom4parisot:
Hello :-)

I'm considering using `tape` for a project requiring IE8 testing, in order to replace `chai`/`mocha`.

Is there any way to achieve doing `before`/`beforeEach`/`after`/`afterEach` to setup fixtures and stubs before running assertions?

Thanks very much :-)

knownasilya:
For others that are interested, there is a package called [redtape](https://www.npmjs.com/package/redtape) that extends `tape` to give you `beforeEach` and `afterEach` functionality.

braddunbar:
> For before()/after() you can just use ordinary tests to do setup and teardown:

So simple… :heart:

sinkingshriek:
@Raynos I just wanna say that, that is some clever piece of code there. My mind was figuratively blown when I finally understood what was going on. :clap:

tomekwi:
@oncletom @Raynos @sinkingshriek you can also [listen to the `end` event](https://github.com/studio-b12/tape-css/blob/0c9285b89b08ecf38d118ecae1beedf8bbe6e875/module/index.js#L57-L81) on `assert`. Benefit: it also works with `assert.plan`.

kuraga:
Any ideas about merging this functionality? From [tapes](https://github.com/scottcorgan/tapes), for instance...

alathon:
Also ran into this. Thanks for the excellent idea @Raynos. I'm still not 100% satisfied with this, as it prevents accessing the Test object to set eg. tape.onFinished(), so that has to be done before you ever call beforeEach (afaik?), f.ex.

I ended up re-implementing it slightly, so that the beforeEach/afterEach methods can take either a function, or an array of functions, as follows:

```javascript
exports.beforeEach = (_test, handlers) => {
  if (typeof handlers === 'function') return _beforeEach(_test, handlers)

  return handlers.reduce((prev, cur, curIdx, arr) => {
    if (curIdx === 0) return prev
    return _beforeEach(prev, cur)
  }, _beforeEach(_test, handlers[0]))
}

exports.afterEach = (_test, handlers) => {
  if (typeof handlers === 'function') return _afterEach(_test, handlers)

  return handlers.reduce((prev, cur, curIdx, arr) => {
    if (curIdx === 0) return prev
    return _afterEach(prev, cur)
  }, _afterEach(_test, handlers[0]))
}

const _beforeEach = (_test, handler) => {
  return function tapish (name, listener) {
[...]

PR #377, opened 2017-07-26
Add support for before and after hook

amrish7:


ljharb:
Rather than creating a fork, have you attempted to PR any of these features into tape itself?

amrish7:
Hi @ljharb The Fork is temporary... I just want to add more features to my fork and do a one shot PR with these features. Moreover, I wasn't sure if the community was open to `before` and `after` hooks on Tape

gabrielcsapo:
the difference between mocha and tape is the non existence of life cycle hooks that causes unit tests to not be pure, isn't that a fundamental philosophy of tape? It is the reason I use it for all of my projects open source or work related.

snuggs:
Well.. @gabrielcsapo doesn't necessarily mean "non pure" just means have to take care of ifecycle within each respective test example. I just use a `before` and a `after` helper that can take parameters perhaps a callback. Just a thought. There's ways to make "pure" just the developer ergonomics aren't there. That said, lifecycle hooks would be nice.

gabrielcsapo:
@snuggs I am just reiterating some design principles that were discussed to differentiate mocha 

- https://github.com/substack/tape/issues/59
- https://github.com/tapjs/node-tap/issues/138

Even @substack said;

```
I vote for no. The economy of using test() for serial execution blocks is very simple and nice. I think a section in the docs about how to beforeEach/afterEach using test() would be just as good without adding more redundant clutter to the API.
```

amrish7:
I do see benefits from lifecycle hooks especially around pulling boiler plate code (like cleaning up spies/mocks) into these hooks thereby making test cases less noisy. 

That being said, I don't want to push my philosophy on these lifecycle hooks onto the community! I could live with a forked version! May be these features gets pulled in at a later stage or it lives outside in a fork for eternity 😄

gabrielcsapo:
@amrish7 I think as a plugin it would make sense, I like the idea for some use cases, but as a base it makes sense from the statements by substack to keep the runner as serial as possible.

snuggs:
@amrish7 this is AWESOME! I let about 5,000 of my closest friends know. We <3 tape!

capture d ecran 2017-07-28 a 11 24 10

Issue #386, opened 2017-09-08
Failed assertions inside anonymous functions get incorrect line numbers


fongandrew:
Consider the following two tests:

```js
test('test 1', t => {
  (function namedFn() {
    t.fail('1'); // Error reported here
  })();
  t.end();
});
```

```js
test('test 2', t => {
  (() => {
    t.fail('2');
  })(); // Error reported here
  t.end();
});
```

On Node v6.11.2, Tape correctly reports that the first test failed at the line with `t.fail('1')`. It incorrect reports that the second test failed at the point where the anonymous function was executed.

The problem appears Issue appears to stem from this code here: https://github.com/substack/tape/blob/5565ef5c965dbd113408b79b30cb0329368e3127/lib/test.js#L221-L253

Basically, what's happening is that Tape is iterating through each line in an error stack trace, using regexes to find and parse the first line that doesn't originate from within Tape's own directory for use as the origin of the error. The problem is that the regex doesn't account for how a stack trace might look for anonymous functions (or more accurately, functions with undefined names).

That is, with the first example above, the relevant stack trace looks like this:

```
Error: exception
    ...
    at namedFn (/Users/username/project/src/test-file.js:5:7)
    at Test.<anonymous> (/Users/username/project/src/test-file.js:6:5)
    ...
```

The stack trace for the second example though looks like this:

```
Error: exception
    ...
    at /Users/username/project/src/test-file.js:12:7
    at Test.<anonymous> (/Users/username/project/src/test-file.js:13:5)
    ...
```

In the first case, the regex picks on the `namedFn` line. In the second case, the regex can't make sense of the line with no function name and so drops down to the `Test.<anonymous>` line to use instead. Note that although the `Test.<anonymous>` function is indeed anonymous, it still has a name by virtue of being set as a property of a `Test` object -- not so for the anonymous function defined in my second example though, so it shows up without any name in the stack trace.

Context: This happened to me because I was trying to wrap the `test` function to automatically do  some setup and teardown for each test called with the wrapper function.  In case this bites anyone else, workaround for my issue was pretty simple -- just use `Object.definePropery(callback, 'name', { value: 'some-name' })` to give a name to the test callback function inside the wrapper.

ljharb:
Thanks, this definitely needs to be fixed in the regex.

That said, in a world with modules, an IIFE should never be necessary, and I'm confident there's a better way to do what you're hoping to achieve - but thanks for reporting this bug!

fongandrew:
[...]

PR #387, opened 2017-09-08
Check added stack trace parts for filename match

fongandrew:
Fixes https://github.com/substack/tape/issues/386. Also refactors regex checking code a little.

One thing I'm not sure about is how to test this. The issue is actually this line here:
https://github.com/fongandrew/tape/blob/2c242ca9715b6b0bfded23fccb522dd414d96612/lib/test.js#L219

Tape ignores stack trace lines that originate from files starting with that `dir` variable. Currently, it means ignoring all stack trace lines from within tape's own package directory. The problem is that tape's own tests also live within its package directory, so the `at` variable is never set on the result object.

We could try changing `dir` to point to just the `lib` directory inside the package. This would result in the `at` being set and we could check it. This should be fine to do, since the stack trace for an assertion error shouldn't run through any file outside of the `lib` directory, but I'm not sure if that could change in the future.

Also, this would result in 25 or so existing tests failing (since their test outputs don't include the `at` variable). It wouldn't be a big deal to update those tests (correct behavior should be including the `at` variable in those test results), but feedback welcome on that.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -225,18 +225,14 @@ Test.prototype._assert = function assert (ok, opts) {
             }
             
             var s = m[1].split(/\s+/);
-            var filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[1]);
-            if (!filem) {
-                filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[2]);
-                
-                if (!filem) {
-                    filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
-
-                    if (!filem) {
-                        continue;
-                    }
-                }
+            var filemRe = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/;
+            var filem;
+            var sIndex;
+            for (sIndex in s.slice(0, 4)) {
+                filem = filemRe.exec(s[sIndex]);
+                if (filem) break;
             }
+            if (! filem) continue;
             
             if (filem[1].slice(0, dir.length) === dir) {
                 continue;
--- a/lib/test.js
+++ b/lib/test.js
@@ -237,13 +237,13 @@ Test.prototype._assert = function assert (ok, opts) {
             if (filem[1].slice(0, dir.length) === dir) {
                 continue;
             }
-            
-            res.functionName = s[0];
+
+            res.functionName = s.length > 1 ? s[0] : '<anonymous>';
             res.file = filem[1];
             res.line = Number(filem[2]);
             if (filem[3]) res.column = filem[3];
             
-            res.at = m[1];
+            res.at = s.length > 1 ? m[1] : '<anonymous> (' + m[1] + ')';
             break;
         }
     }
--- a/lib/test.js
+++ b/lib/test.js
@@ -217,33 +217,65 @@ Test.prototype._assert = function assert (ok, opts) {
         var e = new Error('exception');
         var err = (e.stack || '').split('\n');
         var dir = __dirname + path.sep;
-        
+
         for (var i = 0; i < err.length; i++) {
-            var m = /^[^\s]*\s*\bat\s+(.+)/.exec(err[i]);
+            /*
+                Stack trace lines may resemble one of the following. We need
+                to should correctly extract a function name (if any) and
+                path / line no. for each line.
+            
+                    at myFunction (/path/to/file.js:123:45)
+                    at myFunction (/path/to/file.other-ext:123:45)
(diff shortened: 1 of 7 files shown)


ljharb:
I think that tests kind of need to *not* ignore stack trace lines that originate within the tests directory, so I'd think updating those tests is the right path here (also totally fine to do that in a separate PR, and this can be rebased after that's merged)

fongandrew:
Updated tests are in https://github.com/substack/tape/pull/388. This PR is rebased off of that one.

ljharb:
Please rebase this on latest master when you can :-D

fongandrew:
Rebased!

lib/test.js
```diff
@@ -225,29 +225,25 @@ Test.prototype._assert = function assert (ok, opts) {
...
-                    filem = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/.exec(s[3]);
-
-                    if (!filem) {
-                        continue;
-                    }
-                }
+            var filemRe = /((?:\/|[A-Z]:\\)[^:\s]+:(\d+)(?::(\d+))?)/;
+            var filem;
+            var sIndex;
+            for (sIndex in s.slice(0, 4)) {
```
ljharb: for..in loops should be avoided; this would be better done as a for loop, or as `Array.prototype.some.call` or something

lib/test.js
```diff
@@ -225,29 +225,25 @@ Test.prototype._assert = function assert (ok, opts) {
...
-            
-            res.functionName = s[0];
+
+            res.functionName = s.length > 1 ? s[0] : '<anonymous>';
             res.file = filem[1];
             res.line = Number(filem[2]);
             if (filem[3]) res.column = filem[3];
             
-            res.at = m[1];
+            res.at = s.length > 1 ? m[1] : '<anonymous> (' + m[1] + ')';
```
ljharb: why s.length instead of m.length here?

lib/test.js
fongandrew: `Array.prototype.forEach` would be the cleanest way to do this, but it's not supported in older versions of Node (or older browsers) without polyfill support. Not sure what Tape's polyfill policy is.

I can switch it to some variation of `for (sIndex = 0; sIndex < s.slice(0, 4).length; s++) { ... }` though.

(I'm not sure what expectations Tape makes

lib/test.js
fongandrew: `m` is the result of the regular expression matching something like `at /path/to/file.js:123:10` or `at functionName /path/to/file.js:123:10`. If we're at this point in the code, `m.length` should always be the same, where `m[1]` is the portion of the regex matching everything after `at`.

`s` is `m[1].split(/\s+/)` So if `s` has a length of 1, we know that`m[1]` is just case where there's just a path with no function name.

Although in writing this, it now occurs to me that the original `m[1].split(/\s+/)` code fails to account for directory and file names with spaces ... lemme go fix that.

fongandrew:
Added test for a file name with a space in it. Rewrote the stack trace regex to handle that scenario. New approach also gets rid of the `for ... in` loop.

lib/test.js
ljharb: True; we have the `for-each` package to handle `forEach` but that doesn't handle `some`.

lib/test.js
ljharb: Maybe we could store `s.length > 1` in a nicely named variable, for clarity?

lib/test.js
```diff
@@ -219,35 +219,59 @@ Test.prototype._assert = function assert (ok, opts) {
...
+                at myFunction (/path/to/file.js:123:45)
+                at myFunction (/path/to/file.other-ext:123:45)
+                at myFunction (/path to/file.js:123:45)
+                at myFunction (C:\path\to\file.js:123:45)
+                at myFunction (/path/to/file.js:123)
+                at Test.<anonymous> (/path/to/file.js:123:45)
+                at Test.bound [as run] (/path/to/file.js:123:45)
+                at /path/to/file.js:123:45
+            */
+            var m = new RegExp(
```
ljharb: it would be really nice if we could use a regex literal, and repeat the parts you want to comment, in the comment, instead of using `new RegExp`

lib/test.js
```diff
@@ -219,35 +219,59 @@ Test.prototype._assert = function assert (ok, opts) {
...
+            if (filePath.slice(0, dir.length) === dir) {
                 continue;
             }
+
+            // Function call description may not (just) be a function name.
+            // Try to extract function name by looking at first "word" only.
+            res.functionName = callDescription.split(/s+/)[0]
+            res.file = filePath;
+            res.line = Number(m[3]);
+            if (m[4]) res.column = m[4];
```
ljharb: `Number(m[4])`?

fongandrew:
Comments addressed (amended old commit).

PR #388, opened 2017-09-10
Fix dirname

fongandrew:
See https://github.com/substack/tape/pull/387.

Currently, when setting the `at` property on the results object, tape ignores any stacktrace lines that originate within Tape's package directory, including its own `tests`. This restricts that check to Tape's `lib` directory instead of the entire package directory, and updates tests accordingly.


fongandrew:
Don't think it's a breaking change.

PR #396, opened 2017-09-22
#395 Use default `require.extensions`

### LINKED ISSUE #137: Looks like no unit test is testing the t.error() function

### LINKED ISSUE #395: Tape does not pass `require.extensions` to `resolve`
We are currently using the legacy `.es` ECMAScript file extension. How we are using this is by appending  a `.es` key to `require.extensions`. Have discussed with members of node and this has been converted into a public API since https://github.com/substack/tape/commit/ffa503ad8acb22fe3108748189c0e1888511d8ba which introduced resolve into tape. However, resolve will not use the default `require` `.extensions` see https://github.com/browserify/resolve/blob/5f6e682396eca043624d0a13286b9015b1e6cfce/lib/sync.js#L23 . Already have a commit that merely adds `{ extensions: require.extensions }` to the call to resolve. Please let me know if I should be opening an issue on `tape` (preferred) or in resolve (possibly should do this as well). Since they have a mechanism to change extensions I feel will be a great start here. Will notify resolve team as well. At the current state have no way to alter resolve and no way to alter what tape sends to resolve. Thanks in advance!

Thanks in advance. /cc @brandondees

capture d ecran 2017-09-21 a 17 34 32

snuggs:
Fixes https://github.com/browserify/resolve/issues/137 & Fixes #395 by using default `require.extensions` collection.

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -18,11 +18,12 @@ if (typeof opts.require === 'string') {
 }
 
 opts.require.forEach(function (module) {
+    var options  = { basedir: cwd, extensions: Object.keys(require.extensions) };
     if (module) {
         /* This check ensures we ignore `-r ""`, trailing `-r`, or
          * other silly things the user might (inadvertently) be doing.
          */
-        require(resolveModule(module, { basedir: cwd }));
+        require(resolveModule(module, options));
     }
 });
 


bin/tape
```diff
@@ -18,11 +18,11 @@ if (typeof opts.require === 'string') {
...
-    if (module) {
-      /* This check ensures we ignore `-r ""`, trailing `-r`, or
-       * other silly things the user might (inadvertently) be doing. */
-      require(resolveModule(module, { basedir: cwd }));
-    }
+    var options  = { basedir: cwd, extensions: Object.keys(require.extensions) };
+        
+    /* This check ensures we ignore `-r ""`, trailing `-r`, or
+     * other silly things the user might (inadvertently) be doing. */
+    module && require(resolveModule(module, options));
```
ljharb: Please revert back to the `if (module)` form - abusing value selection operators for control flow isn't a good style.

ljharb:
Actually it just occurred to me why this isn't done. JSON files and binary `.node` files are part of `require.extensions` but would never be tests.

In other words, it'd be fine to make a `tape` command line runner option that allows you to set the `extensions` list (which is passed to `resolve`), but this absolutely couldn't be default behavior.

bin/tape
snuggs: @ljharb certainly! Can you please provide recent documentation of this stance? As a library author and someone who works at the spec level I can almost guarantee that because this may not be the olde hat style. That hat is still old. However I do have strong opinions loosely held. Would really appreciate a link of sorts to update my thinking. 

That said, correcting change now no problem at all for the sake of the repo style. 👍

snuggs:
> In other words, it'd be fine to make a tape command line runner option that allows you to set the extensions list (which is passed to resolve), but this absolutely couldn't be default behavior. @ljharb 

I concur 💯  How can we move forward in the direction of `--extensions` ? This way we can get the resolve issue https://github.com/browserify/resolve/issues/137 closed as per a recommendation in https://github.com/browserify/resolve/issues/137#issuecomment-331353672

bin/tape
snuggs: Corrected in 278c8b569a6bd4da6ca9cf71626be2ac843b7f91

snuggs: FWIW @ljharb I agree this may not be the style of the repo. However "abuse" is a harsh word and in modern ecmascript styles is a very constricting mindset. Coincidentally am currently working with google team to begin converting parts of code in said patterns. https://github.com/webcomponents/custom-elements/pull/102#issuecomment-331528755

To be clear from _User Land_ best practices this is 💯  correct. However as library authors who are often restricted to payload bytesize this is actually a practice that shaves many many bytes off the final bundle. Just a thought I was having. I still concur that pattern is orthogonal to the styles in this repo. Whatever you need from me let me know.

ljharb: Best practices require a very constricting mindset, and library authors should not be concerned with payload size; that's a top-level app concern. A minifier is capable of reducing patterns like this just fine without needing to reduce the quality of the code.

I'd refer you to https://github.com/airbnb/javascript for good documentation of JS styles that tend to largely represent the majority of the community.

Thanks for making the change!

ljharb:
Let's leave this open for the time being; and then would you be willing to open a new PR that adds the `--extensions` option to `tape`?

snuggs:
> Let's leave this open for the time being; and then would you be willing to open a new PR that adds the --extensions option to tape? @ljharb 

It would be my pleasure to champion the PR. Thanks for the great insight and for such an awesome library! 🎉

ljharb:
@snuggs any interest in the `--extensions` PR?

PR #400, opened 2017-10-07
fix stack where actual==falsy

### LINKED ISSUE #399: failure doesn't produce stacks if actual value is falsy
https://github.com/substack/tape/blob/b66f8f80928bb82f0817407880a139ad118daf01/lib/results.js#L160

    test('failureStacks', function(t) {
      t.equal(true, false); // returns a full stack
      t.equal(false, true); // no stack
      t.end();
    });

nhamer:
Fixes #399

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -157,7 +157,7 @@ function encodeResult (res, count) {
         output += inner + 'at: ' + res.at + '\n';
     }
 
-    var actualStack = res.actual && res.actual.stack;
+    var actualStack = res.actual && (typeof res.actual === 'object' || typeof res.actual === 'function') ? res.actual.stack : undefined;
     var errorStack = res.error && res.error.stack;
     var stack = defined(actualStack, errorStack);
     if (stack) {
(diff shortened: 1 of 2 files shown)


lib/results.js
```diff
@@ -157,7 +157,7 @@ function encodeResult (res, count) {
         output += inner + 'at: ' + res.at + '\n';
     }
 
-    var actualStack = res.actual && res.actual.stack;
+    var actualStack = res.actual !== undefined && res.actual.stack;
```
ljharb: this will cause it to throw if `res.actual` is `null`.

If `res.actual` is a primitive (even a truthy primitive), `.stack` will never have a meaningful value.

This check should probably really be `res.actual && (typeof res.actual === 'object' || typeof res.actual === 'function') ? res.actual.stack : null`

nhamer: I'm working slightly retarded here, on windows a lot of the unit tests fail. Will update the pr.

nhamer: Oh, I see. It was working on my machine for the wrong reason. The actual bug is:

when res.actual === false,
res.actual && res.actual.stack === false, not undefined

ljharb: @nhamer fyi, i edited your comment to remove the inappropriate word.

nhamer:
it was the inverse of the existing test; fixed the existing test's assertion to accurately reflect its arguments

PR #403, opened 2017-10-15
Fix https://github.com/substack/tape/issues/222

nhamer:
For reference, a version which doesn't fix the ordering of asserts vs subtests is at https://github.com/nhamer/tape/tree/issue222_early_subtest

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -93,19 +93,11 @@ Test.prototype.test = function (name, opts, cb) {
     this.emit('test', t);
     t.on('prerun', function () {
         self.assertCount++;
-    })
-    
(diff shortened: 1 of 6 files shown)


ljharb:
Can you re-title your PR to be more descriptive about what it does? Which issue it fixes is incidental :-)

nhamer:
Not at the moment - it looks like add-subtest-async.js needs a t.plan(2) - see comment on https://github.com/substack/tape/issues/222#issuecomment-336568413

nhamer:
@ljharb merge issue - I'd added a t.end()

PR #404, opened 2017-10-15
Fix #105: only use one test runner for results, even if multiple streams are created

### LINKED ISSUE #105: Calling createStream() more than once causes weird behavior on the first test.
I have a use-case where I need to create two streams, one in objectMode and one for raw TAP output, however calling createStream() twice has a side-effect where for the very first test in the suite, the state of whether it has ended or not is ignored making t.plan(x) not work properly in the first test.

```
var test = require('tape');

test.createStream(); // all good
test.createStream(); // this extra call causes "starting test 2" to show up before "finishing test 1"

test('test one', function (t) {
  t.plan(1);
  console.log('starting test 1');
  setTimeout( function() {
    console.log('finishing test 1');
    t.pass();
  }, 100);
});

test('test two', function (t) {
   console.log('starting test 2');
   t.end();
});
```

nhamer:
Fixes https://github.com/substack/tape/issues/105

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -24,6 +24,7 @@ function Results () {
     this._stream = through();
     this.tests = [];
     this._only = null;
+    this._isRunning = false;
 }
 
 Results.prototype.createStream = function (opts) {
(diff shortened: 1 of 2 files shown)


nhamer:
(oh, same fix as https://github.com/substack/tape/pull/361, just with a unit test)

billti:
Nice! Seeing as two people have now contributed roughly the same fix to a years old issue, maybe we'll get it fixed soon :-)  (CC @ljharb )

Issue #367, opened 2017-05-24
End test upon failure


justingosan:
Is there a built-in or reliable way to end running the succeeding tests upon failing one of them?

Thanks,
Justin

ljharb:
We have an onFinish callback, but not an onFailure callback.

That seems like a reasonable thing to add.

olalonde:
Yes, something like `mocha --bail` would be nice. 

```
-b, --bail                      bail after first test failure
```

olalonde:
Just found https://github.com/juliangruber/tap-bail

kevinpeno:
@olalonde Sorry if this is a silly question, but how'd you set this up? I tried using is as `tape test/**/*.js | tap-bail` (got a pipe error) and `tape --tap test/**/*.js | tap-bail` and got no output.

Other notes, I'd like to use `blue-tape` and `faucet` with this, but I disabled them for this test.

olalonde:
@kevinpeno you should probably ask on that repo as I just started using it :P my script is `babel-tape-runner test/custom.js | tap-bail | faucet`

lastmjs:
Can this be closed now?

justingosan:
Closed! Thanks everyone!

ljharb:
Fixed in #408.

PR #408, opened 2017-11-13
Add "onFailure" listener to test harness.

johnhenry:
Adds an onFinish listener to test harness as discussed here: https://github.com/substack/tape/issues/367#issuecomment-303785660

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -104,7 +104,10 @@ Results.prototype._watch = function (t) {
         self.count ++;
 
         if (res.ok) self.pass ++
-        else self.fail ++
+        else {
+            self.fail ++;
+            self.emit('fail');
+        }
     });
     
     t.on('test', function (st) { self._watch(st) });
--- a/test/onFailure.js
+++ b/test/onFailure.js
@@ -1,0 +1,21 @@
+var tap = require("tap");
+var tape = require("../").createHarness();
+
+//Because this test passing depends on a failure,
+//we must direct the failing output of the inner test
+var noop = function(){}
+var mockSink = {on:noop, removeListener:noop, emit:noop, end:noop}
+tape.createStream().pipe(mockSink);
+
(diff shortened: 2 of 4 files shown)


lib/results.js
```diff
@@ -104,7 +104,10 @@ Results.prototype._watch = function (t) {
         self.count ++;
 
         if (res.ok) self.pass ++
-        else self.fail ++
+        else {
+            self.emit("fail");
+            self.fail ++;
```
ljharb: could the sync emission fail? it seems like we should increment `self.fail` first.

johnhenry:
@ljharb , thanks for the feedback! I'm still having trouble coming up with a way of testing for the failure event that doesn't cause the test to fail.

test/onFailure.js
```diff
@@ -0,0 +1,13 @@
+//Note:this test currently fails
+var tap = require("tap");
+var tape = require("../");
+const t = tap.test("on failure", {timeout: 1000}, function (tt) {
+    tt.plan(1);
+    tape.onFailure(function() {
```
ljharb: could this add an `onFinish` listener as well that fails?

johnhenry: @ljharb, i don't quite understand your comment.
This test does, in fact, fail. That is, for the "outer test" to pass, the "inner test" must fail, and this failure propagates. I wonder if you, or anyone else might have an idea about how to get around this?

ljharb: Sorry, what I mean is, add something like `tape.onFinish(function () { tt.fail(); })`

johnhenry: @ljharb I couldn't quite get the method that you mentioned to work, though I was able to get a test working after a while. Hopefully that does it. Thanks again for the help and feedback!

lastmjs:
Is anything else needed for this pull request?

johnhenry:
@lastmjs, i think this is ready, but I'm happy to make any changes that would get this into the next release soon.

Issue #416, opened 2018-01-01
JSON Objects Need Better Formatting In Errors


inadarei:
Case in point, this error output is quite useless since the difference is not even shown (it's swallowed under non-extended "{object}" entities):

```
---
    operator: deepEqual
    expected: |-
      { 'h:link': [ { template: { contentType: 'application/x-www-form-urlencoded', fields: { orderNumber: [Object], productCode: [Object], quantity: [Object] } }, uri: 'http://api.x.io/orders/42/items', label: 'Add Item', action: 'append' } ] }
    actual: |-
      { 'h:link': [ { template: { contentType: 'application/x-www-form-urlencoded', fields: { orderNumber: [Object], productCode: [Object], quantity: [Object] } }, uri: 'http://api.x.io/orders/42/items', label: 'Add Item', action: 'append' } ] }
    at: Test.test (/Users/irakli/Dropbox/Projects/node/kokua/test/unit/siren-reverse.js:77:5)
    stack: |-
      Error: Converted properly
```

An actually helpful output, instead, looks like the following:

```
    operator: deepEqual
    expected: |-
      {
  "h:link": [
    {
      "template": {
        "contentType": "application/x-www-form-urlencoded",
        "fields": {
          "orderNumber": {"type": "hidden", "default": "42"},
          "productCode": {"type": "text"},
          "quantity": {"type": "number", "label": "quantity"}
        }
      },
      "uri": "http://api.x.io/orders/42/items",
      "label": "Add Item",
      "action": "append"
    }
  ]
}
    actual: |-
      {
  "h:link": [
    {
      "template": {
        "contentType": "application/x-www-form-urlencoded",
        "fields": {
          "orderNumber": {"type": "hidden", "default": "42"},
          "productCode": {"type": "text"},
          "quantity": {"type": "number", "label": "Quantity"}
        }
      },
      "uri": "http://api.x.io/orders/42/items",
      "label": "Add Item",
      "action": "append"
    }
  ]
}
```

Granted, you can increase the depth of parsing by passing something like `opts.objectPrintDepth = 5` to `test([name], [opts], cb)` but the output is still a crammed-up single string that is utterly unreadable. 

Also passing additional "opts" to every single test() call is significantly ugly. A default for all tests in a suite would be huge help.

inadarei:
PR is forthcoming...

inadarei:
Closing in favor of more focused #418 and #419 per @ljharb suggestion.

PR #417, opened 2018-01-01
Better formatting for large expected/actual objects

inadarei:
Fixes: #416 as a backwards-compatible option (objectPrintDepth = -1 option enables this mode).

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -151,7 +151,8 @@ function encodeResult (res, count) {
         }
         if (typeof res.expected === 'object' &&
             typeof res.actual   === 'object' &&
-            res.objectPrintDepth === -1) {
+            res.objectPrintDepth             &&
+            parseInt(res.objectPrintDepth) === -1) {
 
             var ex = stringify(res.expected);
             var ac = stringify(res.actual);


readme.markdown
```diff
@@ -148,8 +148,10 @@ Available `opts` options are:
 - opts.timeout = 500. Set a timeout for the test, after which it will fail.
   See test.timeoutAfter.
 - opts.objectPrintDepth = 5. Configure max depth of expected / actual object
-  printing.
-  
+  printing. If set to "-1", entire object will be printed as a well-formatted
```
ljharb: Why `-1` when JS has the value `Infinity`?

lib/results.js
```diff
@@ -132,22 +133,34 @@ function encodeResult (res, count) {
...
     if (has(res, 'expected') || has(res, 'actual')) {
-        var ex = inspect(res.expected, {depth: res.objectPrintDepth});
-        var ac = inspect(res.actual, {depth: res.objectPrintDepth});
-        
+        if (process.env.NODE_TAPE_OBJECT_PRINT_DEPTH) {
+            res.objectPrintDepth = process.env.NODE_TAPE_OBJECT_PRINT_DEPTH;
+        }
+        if (typeof res.expected === 'object' &&
+            typeof res.actual   === 'object' &&
+            res.objectPrintDepth             &&
```
ljharb: please do not use spaces for alignment (nor align anything ever, except semantically). There should only be one space around binary operators.

inadarei: ok

readme.markdown
inadarei: The most useful part of the PR is probably the ability to set a global depth via the environmental variable. I don't think environmental variables have value for Infinity, do they?

ljharb: Env vars are always strings; the string "Infinity" would work just fine.

However, I think that this PR seems to be doing two things, and it should only be doing one - can you split the "env var" and the "json stringify" parts into separate PRs?

inadarei: Yeap, makes sense. Will do. Thank you

inadarei:
Part of this is implemented by #420 and the rest with https://www.npmjs.com/package/tap-nirvana, so this should be all good now.

Issue #419, opened 2018-01-02
Better Formatting of Large JSON Objects in expected/actual Reports


inadarei:
Couple Issues with the current formatting:

1. The threshold set implicitly or explicitly may be insufficient, in which case the mis-matched value could be swallowed by an un-expanded portion of `{object}` literal.
2. Even when the threshold is set high-enough, the output is a single-line string that is quite messy and unreadable

Case in point:

```
---
    operator: deepEqual
    expected: |-
      { 'h:link': [ { template: { contentType: 'application/x-www-form-urlencoded', fields: { orderNumber: [Object], productCode: [Object], quantity: [Object] } }, uri: 'http://api.x.io/orders/42/items', label: 'Add Item', action: 'append' } ] }
    actual: |-
      { 'h:link': [ { template: { contentType: 'application/x-www-form-urlencoded', fields: { orderNumber: [Object], productCode: [Object], quantity: [Object] } }, uri: 'http://api.x.io/orders/42/items', label: 'Add Item', action: 'append' } ] }
    at: Test.test (/Users/irakli/Dropbox/Projects/node/kokua/test/unit/siren-reverse.js:77:5)
    stack: |-
      Error: Converted properly
```

can be auto-formatted more legibly as:

```
operator: deepEqual
    expected: |-
      {
  "h:link": [
    {
      "template": {
        "contentType": "application/x-www-form-urlencoded",
        "fields": {
          "orderNumber": {"type": "hidden", "default": "42"},
          "productCode": {"type": "text"},
          "quantity": {"type": "number", "label": "quantity"}
        }
      },
      "uri": "http://api.x.io/orders/42/items",
      "label": "Add Item",
      "action": "append"
    }
  ]
}
    actual: |-
      {
  "h:link": [
    {
      "template": {
        "contentType": "application/x-www-form-urlencoded",
        "fields": {
          "orderNumber": {"type": "hidden", "default": "42"},
          "productCode": {"type": "text"},
          "quantity": {"type": "number", "label": "Quantity"}
        }
      },
      "uri": "http://api.x.io/orders/42/items",
      "label": "Add Item",
      "action": "append"
    }
  ]
}
```

ljharb:
This is partially addressed. with #420.

inadarei:
Correct. Now that #420 has been merged, I was able to implement the rest with https://github.com/inadarei/tap-better-diff ([tap-nirvana](https://www.npmjs.com/package/tap-nirvana) on NPM).

[...]

PR #420, opened 2018-01-02
[New] use `process.env.NODE_TAPE_OBJECT_PRINT_DEPTH` for the default object print depth

inadarei:
Issue #419

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -8,6 +8,7 @@ var trim = require('string.prototype.trim');
 var bind = require('function-bind');
 var forEach = require('for-each');
 var isEnumerable = bind.call(Function.call, Object.prototype.propertyIsEnumerable);
+var toLowerCase = bind.call(Function.call, String.prototype.toLowerCase);
 
 module.exports = Test;
 
@@ -52,11 +53,22 @@ function Test (name_, opts_, cb_) {
     this.pendingCount = 0;
     this._skip = args.opts.skip || false;
     this._timeout = args.opts.timeout;
-    this._objectPrintDepth = args.opts.objectPrintDepth || 5;
     this._plan = undefined;
     this._cb = args.cb;
     this._progeny = [];
     this._ok = true;
+    var depthEnvVar = process.env.NODE_TAPE_OBJECT_PRINT_DEPTH;
+    if (args.opts.objectPrintDepth) {
+        this._objectPrintDepth = args.opts.objectPrintDepth;
+    } else if (depthEnvVar) {
+        if (toLowerCase(depthEnvVar) === 'infinity') {
+            this._objectPrintDepth = Infinity;
+        } else {
+            this._objectPrintDepth = depthEnvVar;
+        }
+    } else {
+        this._objectPrintDepth = 5;
+    }
 
     for (var prop in this) {
         this[prop] = (function bind(self, val) {
--- a/readme.markdown
+++ b/readme.markdown
@@ -148,8 +148,9 @@ Available `opts` options are:
 - opts.timeout = 500. Set a timeout for the test, after which it will fail.
   See test.timeoutAfter.
 - opts.objectPrintDepth = 5. Configure max depth of expected / actual object
-  printing.
-  
+  printing. Environmental variable `NODE_TAPE_OBJECT_PRINT_DEPTH` can set the
+  desired default depth for all tests; locally-set values will take precedence.
+
 If you forget to `t.plan()` out how many assertions you are going to run and you
 don't call `t.end()` explicitly, your test will hang.
 


readme.markdown
```diff
@@ -15,10 +15,10 @@ var test = require('tape');
 
 test('timing test', function (t) {
     t.plan(2);
-
```
ljharb: please revert unrelated whitespace changes.

inadarei: Ha, yeah, I was just looking into how to stop VS Code from adding those :)

readme.markdown
```diff
@@ -148,7 +148,8 @@ Available `opts` options are:
 - opts.timeout = 500. Set a timeout for the test, after which it will fail.
   See test.timeoutAfter.
 - opts.objectPrintDepth = 5. Configure max depth of expected / actual object
-  printing.
+  printing. Environmental variable `NODE_TAPE_OBJECT_PRINT_DEPTH` can set the 
+  desired depth for all tests and overrides any locally-set values.
```
ljharb: should this override? or should the env var merely set the default, so that local intentional overrides still take precedence?

inadarei: Was wondering the same, since good arguments can be made for both approaches. Finally decided to go the current route because:
1. seems more compatible with 12-Factor Apps guidance
2. env vars overriding is how https://www.npmjs.com/package/config works

ljharb: I've no interest in complying with "12-factor apps" by itself; if it isn't the best thing for the project, then it shouldn't be done.

In general, i would agree that an env var should always override an *equivalent setting* - but there is no equivalent for "set every object print depth" in tape. Currently, every one is per-test, and per-test overrides imo should always take precedence.

inadarei: OK, no problem. That makes sense. I was just explaining the logic behind the initial decision, but not doing it that way also makes sense. I will make the change as you requested.

lib/test.js
```diff
@@ -52,11 +52,22 @@ function Test (name_, opts_, cb_) {
...
     this._skip = args.opts.skip || false;
     this._timeout = args.opts.timeout;
-    this._objectPrintDepth = args.opts.objectPrintDepth || 5;
     this._plan = undefined;
     this._cb = args.cb;
     this._progeny = [];
     this._ok = true;
+    var depthEnvVar = 'NODE_TAPE_OBJECT_PRINT_DEPTH';
+    if (args.opts.objectPrintDepth) {
+        this._objectPrintDepth = args.opts.objectPrintDepth
```
ljharb: this line is missing a semicolon

lib/test.js
```diff
@@ -52,11 +52,22 @@ function Test (name_, opts_, cb_) {
...
+    } else if (process.env[depthEnvVar]) {
```
ljharb: instead of storing the key name in a variable, and extracting it multiple times, can we just extract the value into a variable right off the bat?

lib/test.js
```diff
@@ -52,11 +52,22 @@ function Test (name_, opts_, cb_) {
...
+        if (process.env[depthEnvVar].toLowerCase() === 'infinity') {
```
ljharb: we should try to be robust against `delete String.prototype.toLowerCase && delete RegExp.prototype.test` (by caching the prototype method at module level)

lib/test.js
```diff
@@ -52,11 +52,22 @@ function Test (name_, opts_, cb_) {
...
+            this._objectPrintDepth = Infinity;
+        } else {
+            this._objectPrintDepth = parseInt(process.env[depthEnvVar], 10);
```
ljharb: hm, i'm not actually sure about the parseInt here. tape doesn't seem to do any typechecking on the `objectPrintDepth` option - ie, you could pass a non-number or a non-integer and it'd just coerce to a number later. Maybe we should do the same here?

inadarei: ok

lib/test.js
inadarei: Sorry, I am not sure I understand what you are proposing.

lib/test.js
```diff
@@ -56,14 +56,14 @@ function Test (name_, opts_, cb_) {
     this._cb = args.cb;
     this._progeny = [];
     this._ok = true;
-    var depthEnvVar = 'NODE_TAPE_OBJECT_PRINT_DEPTH';
+    var depthEnvVar = process.env['NODE_TAPE_OBJECT_PRINT_DEPTH']
```
ljharb: please always use dot access instead of bracket access in javascript wherever possible.

lib/test.js
```diff
@@ -52,11 +52,22 @@ function Test (name_, opts_, cb_) {
...
-    this._objectPrintDepth = args.opts.objectPrintDepth || 5;
     this._plan = undefined;
     this._cb = args.cb;
     this._progeny = [];
     this._ok = true;
+    var depthEnvVar = process.env.NODE_TAPE_OBJECT_PRINT_DEPTH;
+    if (args.opts.objectPrintDepth) {
+        this._objectPrintDepth = args.opts.objectPrintDepth;
+    } else if (depthEnvVar) {
+        if (depthEnvVar.toLowerCase() === 'infinity') {
```
ljharb: To clarify: we should have `var toLowerCase = String.prototype.toLowerCase;` near the top of the file, and then here, do `toLowerCase.call(depthEnvVar)` instead of assuming that nobody's done `delete String.prototype.toLowerCase` inside their tests.

inadarei:
Thanks for the approval and assitance in getting there. What are the next steps for a merge?

PR #421, opened 2018-01-05
Set property '_isOnly' on t object when test.only

drschwabe:
Did not see any existing way to check this - so I just added this commit which sets t._isOnly = true if inside a test.only function

Useful for doing conditional stuff if running a given test in isolation; for example some database spinup/teardown or  data structure scenario that should be filled in that would otherwise be there if all tests were run.

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -86,6 +86,7 @@ Results.prototype.push = function (t) {
 
 Results.prototype.only = function (t) {
     this._only = t;
+    t._isOnly = true; 
 };
 
 Results.prototype._watch = function (t) {
--- a/test/only6.js
+++ b/test/only6.js
@@ -1,0 +1,6 @@
+var test = require('../');
+
+test.only('only6 _isOnly property', function (t) {
+    t.ok(t._isOnly);
+    t.end();
+});


drschwabe:
If you want to run a series of tape.tests as part of a broader integration test (ie- each test modifying a global state; database; etc) then this property is handy; it lets you keep that integration test / series of tests intact; while preserving the ability to use test.only (which you can use to fill in the necessary state or environment that would have otherwise been performed on previous tests).  Ie: 

```
test('spinup', (t) => {
  db.teardown()
  db.rebuild()
  t.ok()
})
test.only('try feature', (t) => {
  if(t._isOnly) {
    db.teardown()
    db.rebuild() 
  }
  //continue with test
})
```

ljharb:
The proper way to handle that is to run teardown using `test.onFinish` (setup can just be done at module level). That way, it'd run identically for "only" or "all".

drschwabe:
Yeah doing a teardown on test.onFinish() is good.  And like you say pre-setup is can be done  at module level. 

But it's the tests in between that this feature will be most useful for.  For example, saves us from needing this workaround: 

```
//^^ some tests that resulted in 3 users created 

//workaround for not having context on t object: 
testOnly = true 
test.only('test a different feature, assuming 3 users', (t) => {
  if(testOnly) setupThreeUsers()
  //continue test
})

//with this PR: 
test.only('test a different feature, assuming 3 users', (t) => {
  if(t._isOnly) setupThreeUsers()
  //continue test
})

ljharb:
I think this is the wrong solution to your problem.

`.only` is a debugging aid; it's not something your code should be depending on.

If a specific test needs specific setup/teardown, it should specify it as part of the test - whether it's the only one running or not.

drschwabe:
If you have 6 tests that each have a 2 minute spinup/teardown that's 12 minutes of waiting for spinup/teardown. You can save 10 minutes by allowing the tests to share state.  

For debugging, exactly we can use test.only to do a specific test; in which case it should be instructed to run the 2 minute spinup/teardown.   The only way I'm aware to accommodate for the logic there is with the workaround I shared above; not a big deal just an extra line of code to type - as such, I completely understand if you prefer to not include this property. 

Either way, thanks for working through the idea/problem with me.

ljharb:
`specialOnly(test, msg, callback)` and then your `specialOnly` can do that setup/teardown and `test.only` for you?

drschwabe:
Sure that technique would work !

PR #423, opened 2018-01-30
Add less and greater than operators

watson:
If you previously wrote this:

```js
t.ok(foo < 42)
```

It wouldn't tell you the value of `foo` which can be really helpful when debugging.

There were ways around that, either involving creating a custom message containing the value of `foo`:

```js
t.ok(foo < 42, 'should be less than 42 (was ' + foo + ')')
```

Or use the undocumented `extra` property:

```js
t.ok(foo < 42, 'should be less than 42', {actual: foo})
```

Both of these solutions requires a lot of typing, so this PR adds these convenient new operators:

```js
t.lessThan(1, 2) // alias t.lt
t.greaterThan(2, 1) // alias t.gt
t.lessThanOrEqual(1, 1) // alias t.lte
t.greaterThanOrEqual(1, 1) // alias t.gte
```

I didn't find a place where basic operators was tested, so I created a `test/operators-ok.js` file. If that's ok, I suggest that the other operators are tested here as well 😃


watson:
Forgot to mention how it looks in case these new operators fail or pass:

Passing:

```
ok 1 should be less than
ok 2 should be greater than
ok 3 should be less than or equal
ok 4 should be greater than or equal
```

Example of `lessThan` failing:

```
not ok 8 should be less than
  ---
    operator: lessThan
    expected: '< 1'
    actual:   1
    at: Test.<anonymous> (/Users/watson/code/node_modules/tape/foo.js:12:5)
    stack: |-
      Error: should be less than
          at Test.assert [as _assert] (/Users/watson/code/node_modules/tape/lib/test.js:225:54)
          at Test.bound [as _assert] (/Users/watson/code/node_modules/tape/lib/test.js:77:32)
          at Test.lt.Test.lessThan (/Users/watson/code/node_modules/tape/lib/test.js:477:10)
          at Test.bound [as lessThan] (/Users/watson/code/node_modules/tape/lib/test.js:77:32)
          at Test.<anonymous> (/Users/watson/code/node_modules/tape/foo.js:12:5)
          at Test.bound [as _cb] (/Users/watson/code/node_modules/tape/lib/test.js:77:32)
          at Test.run (/Users/watson/code/node_modules/tape/lib/test.js:96:10)
          at Test.bound [as run] (/Users/watson/code/node_modules/tape/lib/test.js:77:32)
          at Immediate.next [as _onImmediate] (/Users/watson/code/node_modules/tape/lib/results.js:71:15)
          at runCallback (timers.js:756:18)
  ...
```

ljharb:
Certainly we should document the "extra" argument, at the least :-)

watson:
@ljharb Comparing numbers is quite common in programming, so IMHO if we can make this experience better, then it's worth it even though it doesn't work for stings. If that's the only issue, I can just update the documentation to say this so that people don't get confused.

What's the use-case for aligning tape with the Node core `assert` module? There's already a few places where this isn't the case, e.g. `t.deepLooseEqual` and `t.notDeepLooseEqual`

watson:
@ljharb ...Actually come to thing of it: This does work for strings - quite well in fact. Actually this works for everything and there's no ambiguity IMHO. If someone wants to ask if an Array is greater than an Object, by all means let them do it. They can do that today by using `t.ok(arr > obj)`. These functions should never try to be clever. All they should do is use the given comparator on the two arguments.

ljharb:
It works for strings only when comparing by ascii character code; this doesn't internationalize properly. To compare strings correctly, one must use `a.localeCompare(b)`.

I totally agree with you that they shouldn't try to be clever; but if someone does `arr > obj`, that's a bug - libraries can't prevent user code from having bugs, but they certainly could try to minimize adding APIs that encourage bugs.

bcomnes:
At a minimum the `extra` property should be documented.

lib/test.js
```diff
@@ -471,6 +471,54 @@ Test.prototype.notDeepLooseEqual
     });
 };
 
+Test.prototype.lt
+= Test.prototype.lessThan
+= function (a, b, msg, extra) {
+    this._assert(a < b, {
+        message : defined(msg, 'should be less than'),
+        operator : 'lessThan',
```
ljharb: should the operator here be the english word, or should it be `<`?

lib/test.js
```diff
@@ -471,6 +471,54 @@ Test.prototype.notDeepLooseEqual
     });
 };
 
+Test.prototype.lt
+= Test.prototype.lessThan
+= function (a, b, msg, extra) {
+    this._assert(a < b, {
+        message : defined(msg, 'should be less than'),
```
ljharb: it seems like this default message should maybe be more like `inspect(a) + ' should be less than ' + inspect(b)`?

watson: I thought of that, but the reason why I ended up not implementing it was because none of the other functions in-line the `b` argument like this. But if you feel it makes sense here, I'll happily update it 😃

watson:
Yep, documenting the `extra` property is a great idea. But I'd say that should be a separate PR just to keep things clean

watson:
@ljharb I guess `a < b` vs `a.localeCompare(b)` is similar to how `t.equal` and `t.deepEqual` behaves. It's documented that `t.equal` uses `===`, so no one expects it to do a deep object comparison. Similar `t.lessThan` uses `<`. At least that's how I see it. But I agree this should be properly documented. If you feel the documentation isn't clear enough on this, I'll happily update it 😃

ljharb:
I think there's still a very open question as to whether this PR should go in at all.

In the meantime, it'd be great if you could make a separate PR that adds documentation for existing functionality :-)

ljharb:
Closing; we're going to stick to the `assert` API here. Thanks for the discussion!

PR #424, opened 2018-02-03
Added yarn installation steps

Ajeey:


gabrielcsapo:
I also don't know the value in telling users how to install a package through a package manager, it seems like if you know how to use yarn, the commands to add any package are pretty thoroughly documented here https://yarnpkg.com/en/docs/usage

PR #433, opened 2018-03-28
Add tap-react-browser

mcnuttandrew:
First things first: Tape is a wonderful library, a++++, thank you for building it. 

I would like to add my own tap reporter type thing to the list of reporters, tap-react-browser. As the name would suggest [tap-react-browser](https://www.npmjs.com/package/tap-react-browser) is a component library for doing in browser testing with react+tape. It enables promisification of tests, but allows you to opt out. It plays nice with embedding components. I wasn't sure if it fit into tap reporters or other, so I went with other? But I welcome any feedback.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -15,10 +15,10 @@ var test = require('tape');
 
 test('timing test', function (t) {
     t.plan(2);
-    
+
     t.equal(typeof Date.now, 'function');
     var start = Date.now();
-    
+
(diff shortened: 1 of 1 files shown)


ljharb:
hmm, adding `fs: 'empty'` shouldn't be needed because it's only required in the `tape` binary, which shouldn't be bundled up.

Thanks tho! "other" seems right.

PR #435, opened 2018-04-17
switch from `deep-equal` to `deep-eql`

### LINKED ISSUE #434: recursion in deepEqual() on object graphs with cycles
I have an object graph with cycles, and I'm testing with `tape`, and it keeps on throwing `RangeError: Maximum call stack size exceeded`.

I traced it back to [this issue](https://github.com/substack/node-deep-equal/issues/19) in `deep-equal`.

It was reported march'15, and it doesn't look like `deep-equal` has received much maintenance, so maybe it's time to consider switching to either [`fast-deep-equal`](https://www.npmjs.com/package/fast-deep-equal) or [`deep-eql`](https://github.com/chaijs/deep-eql), both of which seem to be more popular and more recently active?

Although I didn't see a test-case with circular references in any of those two projects' tests either.

If you let me know how you'd like to address this issue, I can try to help?

mindplay-dk:
Fixes #434

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -1,4 +1,4 @@
-var deepEqual = require('deep-equal');
+var deepEqual = require('deep-eql');
 var defined = require('defined');
 var path = require('path');
 var inherits = require('inherits');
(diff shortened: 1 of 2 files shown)


ljharb:
Closing, per https://github.com/substack/tape/issues/434#issuecomment-383414940

PR #437, opened 2018-05-04
Bugfix function name parsing in stacktrace

marques-work:
@ljharb @substack:

The functionName parsing logic has a typo (since bf5a750937df4920ab9e2e965be02ac69dd43f72):

```javascript
// this will split on one or more occurrences of the letter 's'
res.functionName = callDescription.split(/s+/)[0]
```

Changed it to split on `/\s+/` instead.

Also cleaned up trailing whitespaces in the penultimate commit. That part is optional, but why not?

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -282,7 +282,7 @@ Test.prototype._assert = function assert (ok, opts) {
 
             // Function call description may not (just) be a function name.
             // Try to extract function name by looking at first "word" only.
-            res.functionName = callDescription.split(/s+/)[0]
+            res.functionName = callDescription.split(/\s+/)[0]
             res.file = filePath;
             res.line = Number(m[3]);
             if (m[4]) res.column = Number(m[4]);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -21,11 +21,11 @@ exports = module.exports = (function () {
     var lazyLoad = function () {
         return getHarness().apply(this, arguments);
     };
-
```
ljharb: Please revert these unrelated whitespace changes; that's part of the style of this project.

marques-work: @ljharb done

marques-work: @ljharb although, I'm curious -- why would you keep trailing whitespaces, such as that in https://github.com/substack/tape/blob/7261ccc40b207040f0b2586775ca12164ae93922/lib/test.js#L147 for style?

marques-work:
@ljharb Added tests to validate function name parsing (named and anonymous). LMK if you need anything else.

index.js
ljharb: It's not my decision; it's the style of the repo owner.

Issue #208, opened 2015-11-06
tape-catch


tcurdt:
It's a bit unfortunate that one has to resort to use `tape-catch` to have all test throwing exception to fail (why isn't that the default?). Given that I am using `blue-tape` I can only seem to use either or.

Raynos:
Its a bad default as it makes debugging harder.

a node script should exit when an exception is thrown.

tape is a node program.

yoshuawuyts:
100% agreed on what @Raynos says. If you want to catch expected exceptions use `t.throws`:
```js
const assert = require('assert')
const test = require('tape')

test('an exception', function (t) {
  t.plan(1)
  t.throws(compare.bind(null, 1, 2))
})

function compare (num, num) {
  assert.equal(num, num)
}

**Comment by tcurdt** [comment]
Hm. I had tests throwing exceptions and it was going unnoticed. I just noticed because the amount tests were less. Now maybe that's also because of `blue-tape` or how I am using it.
Exiting on exceptions would be just fine - just not quietly ignoring them.

So maybe this isn't really `tape` issue after all then. Seems like I need to dig deeper.

**Comment by Raynos** [comment]
This might be the unhandled rejection problem with promises

**Comment by robhaswell** [comment]
+1 what is the use-case for reporting a test which raised an exception as passing?

**Comment by ljharb** [comment]
@robhaswell testing what kind of exceptions your API throws and where is absolutely essential.

**Comment by robhaswell** [comment]
@ljharb Absolutely I agree, yet the current behaviour doesn't support that at all. Exceptions raised are silently ignored, how does it help test what is thrown? For that you can use `t.throws(fn)`.

**Comment by ljharb** [comment]
Perhaps inside a promise - but that's how promises work. Otherwise, it will terminate the process.

**Comment by chrisnicola** [comment]
This is definitely related to the fact that Node currently doesn't handle promise rejections. With many tests now being written using async/await this creates basically useless test fail messages.

While this is caused by Node.js currently not handling Promise rejection (in 8 there is a warning that eventually this will just fail but it currently doesn't do that).

As a way to behave inline with expectations for tests, I would suggest that tape should check to see if the test function is `async` or a Promise an then call `.catch()` with a function that simply rethrows the error.

For those dealing with this problem right now, a simple global solution, is to require a test helper that includes the following line:

```
process.on('unhandledRejection', error => {
  // Won't execute
  console.log('unhandledRejection', error.test);
});
```

[...]

PR #441, opened 2018-08-26
Fail a test if its callback returns a promise that rejects

KayleePop:
This is mentioned here https://github.com/substack/tape/issues/208#issuecomment-313244971

The benefit is that you can use async functions as the test callback without unexpected results like hanging with no output in the browser.

If the global variable `Promise` isn't defined, then nothing is changed.

### DIFF
--- a/test/promise_fail.js
+++ b/test/promise_fail.js
@@ -1,0 +1,93 @@
+var tap = require('tap');
+var path = require('path');
+var spawn = require('child_process').spawn;
+var concat = require('concat-stream');
+
+var stripFullStack = require('./common').stripFullStack;
+
+tap.test('callback returning rejected promise should cause that test (and only that test) to fail', function (tt) {
+    tt.plan(1);
+
+    var ps = spawn(process.execPath, [path.join(__dirname, 'promises', 'fail.js')]);
+
+    ps.stdout.pipe(concat(function (rows) {
+        var rowsString = rows.toString('utf8');
+
+        if (/^skip\n$/.test(rowsString)) {
+            return tt.pass('the test file indicated it should be skipped');
+        }
+
+        strippedString = stripFullStack(rowsString);
+
+        // hack for consistency across all versions of node
+        // some versions produce a longer stack trace for some reason
+        // since this doesn't affect the validity of the test, the extra line is removed if present
+        // the regex just removes the lines "at <anonymous>" and "[... stack stripped ...]" if they occur together
+        strippedString = strippedString.replace(/.+at <anonymous>\n.+\[\.\.\. stack stripped \.\.\.\]\n/, '');
+
+        tt.same(strippedString, [
+            'TAP version 13',
+            '# promise',
+            'not ok 1 Error: rejection message',
+            '  ---',
+            '    operator: fail',
+            '    stack: |-',
+            '      Error: Error: rejection message',
+            '          [... stack stripped ...]',
+            '  ...',
+            '# after',
+            'ok 2 should be truthy',
+            '',
+            '1..2',
+            '# tests 2',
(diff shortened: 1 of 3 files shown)


KayleePop:
The build is failing because the stack in the test is slightly different in the different versions of node. It still runs correctly, but the assertion fails because it's not exactly the same TAP output.

Not sure how to deal with that.

test/promises/fail.js
```diff
@@ -0,0 +1,12 @@
+var test = require('../../');
+
+test('promise', function (t) {
```
ljharb: this test should be skipped when `typeof Promise === 'undefined'`

test/promises/subTests.js
```diff
@@ -0,0 +1,13 @@
+var test = require('../../');
+
+test('promise', function (t) {
```
ljharb: also this

test/promises/fail.js
KayleePop: I skipped them using node-tap in promise_fail.js, here https://github.com/substack/tape/pull/441/files#diff-0e44c4bcf1645d05ba9b5c2e44250c26R8

If that variable skip is truthy, then these test files aren't even run. (the builds that don't support promises show the skip in the output)

Should I switch to using test.skip in tape?

ljharb: hm - these tests need to continue to use tap, but it'd be ideal for the skip logic to be in each test file.

KayleePop: It would produce a different (non-failing) TAP output then, which would fail the tap test instead of skipping it.

Should I put it in both? Maybe a comment explaining that they'll be skipped?

ljharb: ah yes, i like putting something in both - a generic "skip" mechanism in the tap test, and a specific one to promises here?

KayleePop: Added a commit for this. It looks at the beginning of the stdout from the test, and if it's "skip", then it passes the test. Each test logs "skip" if `Promise` is undefined.

Is that what you had in mind?

lib/test.js
```diff
@@ -92,8 +93,17 @@ Test.prototype.run = function () {
         this.timeoutAfter(this._timeout);
     }
     this.emit('prerun');
-    this._cb(this);
+    var callbackReturn = this._cb(this);
     this.emit('run');
+
+    if (typeof Promise !== 'undefined') {
```
ljharb: let's ensure Promise is a function, as is Promise.resolve

test/promise_fail.js
```diff
@@ -0,0 +1,85 @@
...
+tap.test('callback returning rejected promise should cause that test (and only that test) to fail', function (tt) {
+    tt.plan(1);
+
+    var ps = spawn(process.execPath, [path.join(__dirname, 'promises', 'fail.js')]);
+
+    ps.stdout.pipe(concat(function (rows) {
+        var rowsString = rows.toString('utf8');
+
+        // if the output starts with skip, then skip this test
+        if (/^skip/.test(rowsString)) {
```
ljharb: should this be `/^skip$/`?

KayleePop: Added a commit that only runs the test if promises are supported.

Before it would still run the test and print the TAP output after 'skip'

lib/test.js
KayleePop: done. Also changed in lib/test.js

ljharb:
@KayleePop this needs to be rebased now that #472 has landed. I'm hoping the tests can be preserved :-)

KayleePop:
Rebased!

I just left master's implementation alone and only rebased the tests in. The only difference was that this branch checked that `Promise.resolve` was a function too instead of just `Promise`

PR #443, opened 2018-09-10
Remove broken image from readme

styfle:
This image appears to be broken so I removed it


readme.markdown
```diff
@@ -6,8 +6,6 @@ tap-producing test harness for node and browsers
 
 [![build status](https://secure.travis-ci.org/substack/tape.svg?branch=master)](http://travis-ci.org/substack/tape)
 
-![tape](http://substack.net/images/tape_drive.png)
```
ljharb: let's update this to https://web.archive.org/web/20170612184731if_/http://substack.net/images/tape_drive.png instead

styfle: Fixed 🔧

PR #451, opened 2018-12-13
Add -i flag to ignore tests inside node_module directories

### LINKED ISSUE #428: Ability to ignore node_modules or files in .gitignore by default
Referencing issue #427, I am working on a project where I'm co-locating tests with their module.

i.e.
```
--/path/to/module.js
--/path/to/module.spec.js
```

I'm currently targeting these spec files with the function `tape '{,!(node_modules)/**/}*.spec.js' | tap-spec`. The glob I've written is somewhat convoluted and slightly difficult to read and it does seem like perhaps the default behaviour for the tape-cli is that it should exclude the `node_modules` directory or respect your `.gitignore` file.

Within the glob module that tape uses, there is an `ignore` option to exclude certain directories and files. If there are legitimate reasons for not excluding node_modules and/or gitignore by default, perhaps this could be achieved behind a flag.

Thoughts?

kepikoi:
This will partly resolve #428 by adding the -i flag to ignore any tests inside node_module directories to allow co-locating tests with their modules. Running `tape **/*.spec.js -i` in the project directory won't execute tests inside its npm packages that may break the tape run due different testing frameworks etc.

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -4,13 +4,15 @@
 
 var resolveModule = require('resolve').sync;
 var resolvePath = require('path').resolve;
+var readFileSync = require('fs').readFileSync;
 var parseOpts = require('minimist');
 var glob = require('glob');
+var ignore = require('dotignore');
 
 var opts = parseOpts(process.argv.slice(2), {
-    alias: { r: 'require' },
-    string: 'require',
-    default: { r: [] }
+    alias: { r: 'require', i: 'ignore' },
+    string: ['require', 'ignore'],
+    default: { r: [], i: null }
 });
 
 var cwd = process.cwd();
@@ -29,6 +31,16 @@ opts.require.forEach(function (module) {
     }
 });
 
+if (typeof opts.ignore === 'string') {
+    try {
+        var ignoreStr = readFileSync(resolvePath(cwd, opts.ignore || '.gitignore'), 'utf-8');
+    } catch (e) {
+        console.error(e.message);
+        process.exit(2);
+    }
+    var matcher = ignore.createMatcher(ignoreStr);
+}
+
 opts._.forEach(function (arg) {
     // If glob does not match, `files` will be an empty array.
     // Note: `glob.sync` may throw an error and crash the node process.
@@ -38,7 +50,7 @@ opts._.forEach(function (arg) {
         throw new TypeError('unknown error: glob.sync did not return an array or throw. Please report this.');
     }
 
-    files.forEach(function (file) {
+    files.filter(function (file) { return !matcher || !matcher.shouldIgnore(file); }).forEach(function (file) {
         require(resolvePath(cwd, file));
     });
 });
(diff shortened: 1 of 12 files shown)


kepikoi:
Sorry, forgot to change the branch and https://github.com/substack/tape/pull/451/commits/2fc5d1f8df3408b95a00b075a895415749178f52 https://github.com/substack/tape/pull/451/commits/0ba37b04be692ddbded33f86f1175ba6c1550a3a creeped accidentally in. I undid these changes in the following commits

kepikoi:
I'm not quite sure - do you want me to remove the i flag and make it ignore node_modules by default? Or are you saying the i flag makes only sense when excluding .gitignore entries is implemented?

ljharb:
I think I’m saying a few things:
1. a generic “ignore” flag would have to take explicit paths, or a path to an ignore file, and not default to “node_modules”
1. the root issue is that tape is including paths it shouldn’t be testing - namely, anything not committee to the repo. node_modules is just a symptom of this.
1. I think excluding gitignored files (under an option) is the best use of time and effort to address this problem

Are you willing to update this PR to implement https://github.com/substack/tape/issues/428#issuecomment-367939098 ?

kepikoi:
Ok, I will update the pr to exclude gitignored files when using `-i` and let it ignore anything under node_modules by default.

bin/tape
```diff
@@ -6,23 +6,23 @@ var parseOpts = require('minimist');
...
     opts.require = [opts.require];
 }
 
-opts.require.forEach(function(module) {
+opts.require.forEach(function (module) {
     if (module) {
         /* This check ensures we ignore `-r ""`, trailing `-r`, or
          * other silly things the user might (inadvertently) be doing.
          */
-        require(resolveModule(module, { basedir: cwd }));
```
ljharb: please revert this spacing change (and the others)

readme.markdown
```diff
@@ -84,6 +84,14 @@ $ tape -r ./my/local/module tests/**/*.js
 
 Please note that all modules loaded using the `-r` flag will run *before* any tests, regardless of when they are specified. For example, `tape -r a b -r c` will actually load `a` and `c` *before* loading `b`, since they are flagged as required modules.
 
+## Co-locating tests
+
+When co-locating tests with their modules adding the `-i` flag will allow to ignore any tests that reside inside /node_modules/ directories. This will allow tape processing tests recursively in the project:
```
ljharb: I don't recommend colocation at all, so I don't want that mentioned in the readme.

r0mflip:
+1 for "ignoring things that are gitignored"

ljharb:
@kepikoi are you interested in completing this PR?

kepikoi:
Sorry, no capacity rn for `ignoring things that are gitignored`. Closing the PR

ljharb:
Replacing this with #492.

PR #454, opened 2018-12-22
[Fix] notEqual and notDeepEqual show "expected" value on failure.

### LINKED ISSUE #453: Error messages for notEqual and notDeepEqual never display the expected value.
Any failing tests with `t.notEqual` or `t.notDeepEqual` always show the expected value as `undefined` because they return a result with a `notExpected` property (which is totally ignored) instead of the `expected` property which `_assert` actually looks at.

JoshuaGrams:
Closes #453.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -402,7 +402,7 @@ function notEqual(a, b, msg, extra) {
         message : defined(msg, 'should not be equal'),
         operator : 'notEqual',
         actual : a,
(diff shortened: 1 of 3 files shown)


JoshuaGrams:
OK, try this. But you had better check over my code: I just copied `test/deep-equal-failure.js` and adapted it: I'm not sure I fully understand how it works.

PR #462, opened 2019-03-03
Removed never-read inErrorState from index.js

### LINKED ISSUE #461: Unused variable in index.js: inErrorState
I just happened to notice this poking through `index.js` 😄

JoshuaKGoldberg:
Fixes #461.


ljharb:
This seems to have been added in 9c60d320 / #127, and then not properly cleaned up in e3f9d0193bdf2f8ce03912959fb762a56e908426 / #138.

(note, for future reference, there's no need to create an issue for stuff like this :-) )

Issue #465, opened 2019-04-21
Problem with using stream events


r0mflip:
The stream created by `tape.createStream()` is closed on process exit. Until the process exits the `close` event on the stream won't fire. This is causing issues with tap consumer programs like `tap-spec` and `tap-out` where they cannot hook into the `end` or `close` event of the stream. 

Consider this where `add` pases and `multiply` fails:
``` js
const test = require('tape');
const tapSpec = require('tap-spec');
const {add, divide} = require('../math.js');

// piping programatically
test.createStream()
  .pipe(tapSpec())
  .pipe(process.stdout);

test('should add', t => {
  t.equal(add(3, 4), 7);
  t.end();
});

test('should divide', t => {
  t.equal(divide(8, 2), 4);
  t.end();
});
```
I get the output:
```
> node test\test.js

  should add

    √ should be equal

  should divide


    × should be equal
    ------------------
      operator: equal
      expected: 4
      actual:   6
      at: Test.t (D:\dev\projects\testing\tap-spec-test\test\test.js:15:5)
      stack: |-
  Error: should be equal
  at Test.assert [as _assert] ...
```

Whereas when I use `tap-spec` from command line (removing programatic pipe) `node test/test.js | tap-spec` I get:
```
> node test/test.js | tap-spec

  should add

    √ should be equal

  should divide

    × should be equal
    ------------------
      operator: equal
      expected: 4
      actual:   6
      at: Test.t (D:\dev\projects\testing\tap-spec-test\test\test.js:15:5)
      stack: |-
  Error: should be equal
  at Test.assert [as _assert] ...

  Failed Tests: There was 1 failure

    should divide

      × should be equal

  total:     2
  passing:   1
  failing:   1
  duration:  176ms
```

The results in the bottom are obtained by `tap-out` on the `close` event of the stream. When run the `close` event on the stream won't fire until before the whole process exits, at which point `tap-out` stops working. In command line usage the stream `close` event fires as it is from `stdin` and still the `tap-out` or `tap-spec` process would be alive.

Is there a solution to this problem or should I be just pipe my tap output to other programs.

ljharb:
Is there perhaps something the command line code for tap-spec is doing that your programmatic version isn’t?

r0mflip:
I've checked and experimented, its totally the same. The only thing is when and how the stream ends.

[...]

PR #466, opened 2019-04-29
Add `opts.autoclose` in `test.createStream()` #465

r0mflip:
> Tries to solve #465 

### Problem
The stream created inside `results` was being closed only when the process was exiting. Changed the API to introduce `opts.autoclose` that is provisioned over `canEmitExit` so that results get closed when the results are `done`. In normal usage(from command line) harness closes on process end.

`autoclose` is required beacause `through`s queue method is storing the results untill the end where it is `end`ed only when harness is closed.

### Normal usage
``` js
const tapSpec = require('tap-spec');
const {add, divide} = require('../index.js');
const test = require('./tape/index.js');

// piping programatically
test.createStream()
  .pipe(tapSpec())
  .pipe(process.stdout);

test('should add', t => {
  t.equal(add(3, 4), 7);
  t.end();
});

test('should divide', t => {
  t.equal(divide(8, 2), 4);
  t.end();
});
```
we get the output:
```
> node test/test.js

  should add

    √ should be equal

  should divide

    √ should be equal
```
### `opts.autoclose` set to true
``` js
test.createStream({
  autoclose: true,
})
  .pipe(tapSpec())
  .pipe(process.stdout);
```
we get
```
> node test/test.js

  should add

    √ should be equal

  should divide

    √ should be equal

  total:     2
  passing:   2
  duration:  43ms
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,7 +30,7 @@ exports = module.exports = (function () {
         if (!opts) opts = {};
         if (!harness) {
             var output = through();
-            getHarness({ stream: output, objectMode: opts.objectMode });
+            getHarness({ stream: output, objectMode: opts.objectMode, autoclose: opts.autoclose });
             return output;
         }
         return harness.createStream(opts);
@@ -50,7 +50,7 @@ exports = module.exports = (function () {
 
     function getHarness(opts) {
         if (!opts) opts = {};
-        opts.autoclose = !canEmitExit;
+        opts.autoclose = opts.autoclose || !canEmitExit;
         if (!harness) harness = createExitHarness(opts);
         return harness;
     }
@@ -106,7 +106,7 @@ function createHarness(conf_) {
     if (!conf_) conf_ = {};
     var results = createResult();
     if (conf_.autoclose !== false) {
-        results.once('done', function () { results.close() });
+        results.once('done', function () { if (!results.closed) results.close() });
     }
 
     var test = function (name, conf, cb) {
@@ -151,7 +151,7 @@ function createHarness(conf_) {
     };
     test._exitCode = 0;
 
-    test.close = function () { results.close() };
+    test.close = function () { if (!results.closed) results.close() };
 
     return test;
 }
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@ exports = module.exports = (function () {
 
     function getHarness(opts) {
         if (!opts) opts = {};
-        opts.autoclose = opts.autoclose || !canEmitExit;
+        opts.autoclose = typeof opts.autoclose === 'boolean' ? opts.autoclose : !canEmitExit;
         if (!harness) harness = createExitHarness(opts);
         return harness;
     }


index.js
```diff
@@ -50,7 +50,7 @@ exports = module.exports = (function () {
 
     function getHarness(opts) {
         if (!opts) opts = {};
-        opts.autoclose = !canEmitExit;
+        opts.autoclose = opts.autoclose || !canEmitExit;
```
ljharb: so if `autoclose` is `true`, this will be `true`, but if it's `false`, this won't be guaranteed to be false.

should this instead be?:
```suggestion
        opts.autoclose = typeof opts.autoclose === 'boolean' ? opts.autoclose : !canEmitExit;
```
?
or instead should it be:
```suggestion
        opts.autoclose = canEmitExit ? opts.autoclose : false;
```

r0mflip: `canEmitExit` is based on the environment, but `opts.autoclose` is meant to close the stream before exit whatever the environment be. (The first suggestion is what I'am going for)

What are your thoughts on `canEmitExit`? Should we leave it alone or close the stream once the tests are `done` whatever the environment maybe?

ljharb: I'm a bit unclear still, streams aren't where my knowledge lies :-) can you explain a bit more?

r0mflip: `canEmitExit` is used to know if `tape` is running in browser or in a process like node. If inside a browser the stream that spits out the results/output is closed when the "results end", this is done when the `done` event is fired on the results.

But in environments like node the steam is closed when the process ends. This is to facilitate the use of `tape test/*.js` usage (what I think is). But when this is used the last 3 lines for the output (the one which shows the aggregated resuls) which I was trying won't be spit out when I use `tape` as a module. When run through `tape test/*.js` the stream spits those 3 lines to `stdout`, thus the output is preserved. But when required as a module and a stream is created manually, the output is contained inside the process, when it ends the output is gone because the stream is just a variable that died without spitting it to `stdout`. This is delayed by the `resumer` package. It waits to put the chunks out onto the target stream.

My actual suggestion is to actually remove `canEmitExit` totally and to just close the stream right away. But I think it might cause other problems.

ljharb: What other problems would it solve?

ljharb: Are there circumstances where someone might want the process to exit immediately (initiated by user code) and tape would still need to suppress the final output?

r0mflip: > What other problems would it cause?

I don't think it would cause any

> Are there circumstances where someone might want the process to exit immediately (initiated by user code) and tape would still need to suppress the final output?

I don't think so, if the process exits normally then all the other tests are reported as "exited without ending", when it ends abnormally then nothing extra is printed, the whole process just quits. In both cases the final output is not shown 😟

ljharb: If no tests break, and there's no obvious problems caused by removing `canEmitExit`, we might as well remove it?

r0mflip: Yes, and also I've spotted a few unused variables lying around that have a purpose but are not used at all just like `canEmitExit`. I wonder, why actually they were written.

ljharb: ¯\\\_(ツ)_/¯ it predates my involvement in the project.

r0mflip: I'll see what I can do. I'll try to remove `canEmitExit`, and test it, Add tests for the current situation and remove the `opts.autoclose`

r0mflip: I had a good sleep and after looking around it looks like I was wrong. `canEmitExit` is there so that the process exists with the correct exit code. Looks like we need the explicit `autoclose` for that one specific purpose.

r0mflip: Reporters like `tap-mocha-reporter` work fine without `autoclose`, it consumes the stream quickly and directly spits it to `stdout` only `tap-spec` has this problem because it uses it's own `tap-out` parser which tries to read the stream in a different way.

r0mflip: Alas!! I've found a way where `tap-out` could work just like `tap-mocha-reporter`. We don't need to implement `autoclose` anymore, it depends on the reporter now.

We can ignore this PR now I think 😀

ljharb: Thanks for the update!

ljharb:
Closing per https://github.com/substack/tape/pull/466#discussion_r283122484, but happy to reopen if there's still a need!

PR #470, opened 2019-05-23
[fix] don't consider 'ok' of todo tests in exit code

### LINKED ISSUE #469: {todo: true} passes the test but cli exits with 1
```js
$ npx tape test.js; echo Exit was $? <------------- will print the exit status
TAP version 13
# (anonymous)
not ok 1 should be equal # TODO
  ---
    operator: equal
    expected: false
    actual:   true
    at: Test.t (./test.js:5:4)
  ...

1..1
# tests 1
# pass  1

# ok

Exit was 1 <------------- notice this
```

Smallest repro:

```js
const test = require('tape');

test({todo: true}, t => {
	t.equal(true, false);
	t.end();
});
```

r0mflip:
Fixes #469 

`result.ok` not being considered towards exitcode if `result.todo` is true

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,7 +118,7 @@ function createHarness(conf_) {
                 inspectCode(st_);
             });
             st.on('result', function (r) {
-                if (!r.ok && typeof r !== 'string') test._exitCode = 1
+                if (!r.todo && !r.ok && typeof r !== 'string') test._exitCode = 1
             });
         })(t);
 
--- a/test/todo_single.js
+++ b/test/todo_single.js
@@ -1,0 +1,37 @@
+var tap = require('tap');
+var tape = require('../');
+var concat = require('concat-stream');
+
+var common = require('./common');
+var stripFullStack = common.stripFullStack;
+
+tap.test('tape todo test', function (assert) {
+    var test = tape.createHarness({ exit: false });
+    assert.plan(1);
+
+    test.createStream().pipe(concat(function (body) {
+        assert.equal(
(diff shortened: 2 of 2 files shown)


bfred-it:
Thank you! Hope this is merged soon

bfred-it:
This sadly doesn't fix the issue. The test added does not check the exit code like these tests do https://github.com/substack/tape/blob/master/test/exit.js

ljharb:
A PR that does would be appreciated.

bfred-it:
I looked into the code and I couldn’t find any path that changed the exit code. It seems that the error automatically changes it.

r0mflip:
I will try to add other tests.

> I looked into the code and I couldn’t find any path that changed the exit code. It seems that the error automatically changes it.

I sorry, I don't understand.

bfred-it:
There are a few mentions of `test._exitCode = 1` but even if I disable them all, the exit code is still 1, so there's something else that sets it and I can't find it.

r0mflip:
Tried to add tests in #471

r0mflip:
> @bfred-it There are a few mentions of test._exitCode = 1 but even if I delete them all, the exit code is still 1, so there's something else that sets it and I can't find it.

After the fix #470 I don't get any of those problems. Can you please confirm again, I've tested it and I get the right exit code.

bfred-it:
I tried the latest on npm and I get those errors, even after deleting node_modules

Issue #358, opened 2017-04-14
Test using async/await does not end


mkermani144:
I looked at many issues in this repo, such as #222, #262, etc. but I still don't know why the following test does not end:
```
test('connection to database', async (assert) => {
  const res = await connectToDatabase(mongoose);
  assert.equal(res, 0, 'It should connect to the database without errors');
  assert.end();
});
```
when `connectToDatabase` is an async function.  

When I run the test, I get the following result:

```
# connection to database
  ok 1 It should connect to the database without errors
```
but the test does not finish. Shouldn't `assert.end()` finish the test after promise resolution? What is the reason behind this behavior?

ljharb:
You can't really pass an async function to tape because tape does not do anything with the promise it returns.

You might have slightly more success though using `assert.plan(1)` prior to the first `await`.

mkermani144:
@ljharb 

> You can't really pass an async function to tape because tape does not do anything with the promise it returns.

I don't understand. Why the test passes, but `assert.end()` does not work?

> You might have slightly more success though using assert.plan(1) prior to the first await.

I tried that, but the problem didn't resolve.

mkermani144:
Furthermore, if I add more async tests, i.e.: 
```
test('async test 1', async (assert) => {
  const res = await someAsyncCall();
  assert.equal(res, 0);
  assert.end();
});
test('async test 2', async (assert) => {
  const res = await anotherAsyncCall();
  assert.equal(res, 0);
  assert.end();
});
test('async test 3', async (assert) => {
  const res = await yetAnotherAsyncCall();
  assert.equal(res, 0);
  assert.end();
});
```
only the last one doesn't finish. The others pass and finish correctly.

ljharb:
An async function is just sugar for a function that returns a promise.

Tape does not do anything special with promises.

Without a synchronous call to `.plan()`, tape can't know when your test is expected to end, so having multiple async tests like this can confuse it.

You need to only pass a *function* to tape, not an async function; unless you're willing to jump through all the hoops required, which includes calling .plan before the first await, in every test.

mkermani144:
@ljharb As I said, using `.plan(1)` doesn't help me. Even if I call `.plan(1)` before each of the `await`s, I still get the same result.
I think this issue should be closed, however, as `async/await` is not supported by Tape (currently).

ljharb:
Ah - the likely issue is that one of your tests has a rejected promise, and that's causing the `end` never to be reached. This pattern should work:
```js
[...]

Issue #262, opened 2016-02-18
`Promise` support


Kl0tl:
Would be great to support returning a promise to end the test when the promise is resolved.

+ It would play nice with future extensions of the language, allowing one to call `tape` with Async Functions to transparently return a promise.
+ It would eliminate some otherwise unfixable ordering issues with nested tests.
+ Rewriting

```js
var test = require('tape');

test('first', function (t) {

  setTimeout(function () {
    t.ok(1, 'first test');
    t.end();
  }, 200);

  t.test('second', function (t) {
    t.ok(1, 'second test');
    t.end();
  });
});

test('third', function (t) {
  setTimeout(function () {
    t.ok(1, 'third test');
    t.end();
  }, 100);
});
```

to (assuming a function `delay` returning a `Promise` resolved after `t` milliseconds)

```js
var test = require('tape');

test('first', async function (t) {

  await delay(200);

  t.ok(1, 'first test');

  t.test('second', async function (t) {
    t.ok(1, 'second test');
  });
});

test('third', async function (t) {
  await delay(100);
  t.ok(1, 'third test');
});
```

will preserve the expected order (see https://github.com/substack/tape/issues/222).

---

Looking at [blue-tape](https://github.com/spion/blue-tape/blob/master/blue-tape.js#L7-L34) it should be doable with something along the lines of

```js
var ret = this._cb(this);
var self = this;
if (ret && typeof ret.then === 'function') {
  ret.then(function () {
    self.end();
  }, function (err) {
    nextTick(function () {
      throw err;
    });
  });
}
```

Note that it would not require a `Promise` library nor any configuration. Users wanting to take advantage of this feature would need to provide their own `Promise` implementation if necessary anyway.

ljharb:
From https://github.com/substack/tape/issues/222#issuecomment-185825192:
>
Supporting that is tricky because we'd have to either a) assume Promise was available, which forces the user to know they need to shim it, b) allow explicit injection of `Promise` so they can pass bluebird or native Promises or whatever as they like (like `.setPromiseConstructor()` or something), or c) add a Promise lib as a dependency.
>
Solution a is dangerous, b is awkward, and c is heavyweight.

The problem is that `tape` would need to be able to wrap test return values in `Promise.resolve` since it's a bad idea to do an "is this a promise" check - which means `tape` needs to be able to know that `Promise` is available.

Kl0tl:
[...]

PR #472, opened 2019-06-20
[Breaking] support passing in an async function for the test callback

Raynos:
This PR adds rudimentary support for async/await in a fashion
that breaks back compat as little as possible.

Currently tape has no opinions about exception handling.
If you throw in a test block it will go straight to the
uncaught handler and exit the process.

However newer versions of node & v8 added async functions.
These functions when thrown return a rejected promise.

Currently node logs unhandled rejections with a warning
and in future versions of node, unhandled rejections may get
forwarded to uncaught exceptions.

This patch modifies tape so that an unhandled rejection will be 
marked as a TAP failure and fails the tests with exit code 1.

Previously on master it would log to STDERR and then continue
onwards running the rests of the tests.

For example :

```js
var tape = require('tape')

tape('my test', function (t) {
  // this causes the tests to fail with exit code 1
  t.oktypo(true, 'my bad')
})

tape('my async test', async function (t) {
  // on master : this causes the tests to pass with exit code 0 
  //     and STDERR noise about unhandled rejections
  // in this PR : this causes the tests to fail with exit code 1
  t.oktypo(true, 'my bad')
})
```

This patch is optional, there is a different work around you can do

```js
process.on('unhandledRejection', function (err) { throw err })

var tape = require('tape')

tape('my async test', async function (t) {
  // on master : this causes the tests to fail with exit code 1
  t.oktypo(true, 'my bad')
})
```

But adding that one liner to the top of every test file 
can get quite annoying and EASY to forget :(

Back compat is broken for some cases, only when the user of `tape` mixes promises & callbacks & async functions in a single `test` block. If the async function returns before you call `t.end()` ( aka you do not `await` the code that calls `t.end()`

However no back compat breaks for users that

 - do not use async/await
 - do not use promises
 - uses async/await but only with promises
 - uses promises and no callbacks
 - uses async/await, promises & callbacks but always does `await util.promisify((cb) { ...; t.end() }`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,4 +8,12 @@
             "named": "never",
         }],
     },
+    "overrides": [
+        {
+            "files": ["test/async-await/*"],
+            "parserOptions": {
+                "ecmaVersion": 2017,
+            },
+        },
+    ],
 }
--- a/lib/test.js
+++ b/lib/test.js
@@ -91,7 +91,23 @@ Test.prototype.run = function () {
     if (this._timeout != null) {
         this.timeoutAfter(this._timeout);
     }
-    this._cb(this);
+
+    var callbackReturn = this._cb(this);
+
+    if (
+        typeof Promise === 'function' &&
+        callbackReturn &&
+        typeof callbackReturn.then === 'function' &&
+        typeof callbackReturn.catch === 'function'
+    ) {
+        callbackReturn.catch(function onError(err) {
+            nextTick(function rethrowError() {
+                throw err
+            })
+        })
+        return
+    }
+
     this.emit('run');
 };
 
--- a/lib/test.js
+++ b/lib/test.js
@@ -97,15 +97,18 @@ Test.prototype.run = function () {
     if (
         typeof Promise === 'function' &&
         callbackReturn &&
-        typeof callbackReturn.then === 'function' &&
-        typeof callbackReturn.catch === 'function'
+        typeof callbackReturn.then === 'function'
     ) {
-        callbackReturn.catch(function onError(err) {
-            nextTick(function rethrowError() {
-                throw err
-            })
-        })
-        return
+        var self = this;
+        Promise.resolve(callbackReturn).then(function onResolve() {
+            if (!self.calledEnd) {
+                self.end();
+            }
+        }).catch(function onError(err) {
+            self.fail(err);
+            self.end();
+        });
+        return;
     }
 
     this.emit('run');
--- a/test/async-await.js
+++ b/test/async-await.js
@@ -128,11 +128,10 @@ tap.test('async5', function (t) {
             'ok 4 after request',
             'ok 5 should be equal',
             'ok 6 should be equal',
-            'ok 7 undefined',
             '',
-            '1..7',
-            '# tests 7',
-            '# pass  7',
+            '1..6',
+            '# tests 6',
+            '# pass  6',
             '',
             '# ok'
         ].join('\n') + '\n\n');
@@ -190,7 +189,20 @@ tap.test('async-error', function (t) {
             'TAP version 13',
             '# async-error',
             'ok 1 before throw',
-            ''
+            'not ok 2 Error: oopsie',
+            '  ---',
+            '    operator: fail',
+            '    stack: |-',
+            '      Error: Error: oopsie',
+            '          [... stack stripped ...]',
+            '  ...',
+            '',
+            '1..2',
+            '# tests 2',
+            '# pass  1',
+            '# fail  1',
+            '',
+            '',
         ].join('\n'));
         t.same(r.exitCode, 1);
 
@@ -203,18 +215,7 @@ tap.test('async-error', function (t) {
         });
         stderr = lines.join('\n');
 
-        t.same(stripFullStack(stderr), [
-            '$TAPE/lib/test.js:106',
-            '                throw err',
-            '                ^',
-            '',
-            'Error: oopsie',
-            '    at Test.myTest ($TEST/async-await/async-error.js:$LINE:$COL)',
-            '    at Test.bound [as _cb] ($TAPE/lib/test.js:$LINE:$COL)',
-            '    at Test.run ($TAPE/lib/test.js:$LINE:$COL)',
-            '    at Test.bound [as run] ($TAPE/lib/test.js:$LINE:$COL)',
-            ''
-        ].join('\n'));
+        t.same(stderr, '');
         t.end();
     });
 });
--- a/test/async-await.js
+++ b/test/async-await.js
@@ -119,7 +119,7 @@ tap.test('async4', function (t) {
 
 tap.test('async5', function (t) {
     runProgram('async-await', 'async5.js', function (r) {
-        t.same(r.stdout.toString('utf8'), [
+        t.same(stripFullStack(r.stdout.toString('utf8')), [
             'TAP version 13',
             '# async5',
             'ok 1 before server',
@@ -128,14 +128,24 @@ tap.test('async5', function (t) {
             'ok 4 after request',
             'ok 5 should be equal',
             'ok 6 should be equal',
+            'ok 7 undefined',
+            'not ok 8 .end() called twice',
+            '  ---',
+            '    operator: fail',
+            '    at: Server.<anonymous> ($TEST/async-await/async5.js:$LINE:$COL)',
+            '    stack: |-',
+            '      Error: .end() called twice',
+            '          [... stack stripped ...]',
+            '          at Server.<anonymous> ($TEST/async-await/async5.js:$LINE:$COL)',
+            '          [... stack stripped ...]',
+            '  ...',
             '',
-            '1..6',
-            '# tests 6',
-            '# pass  6',
-            '',
-            '# ok'
+            '1..8',
+            '# tests 8',
+            '# pass  7',
+            '# fail  1'
         ].join('\n') + '\n\n');
-        t.same(r.exitCode, 0);
+        t.same(r.exitCode, 1);
         t.same(r.stderr.toString('utf8'), '');
         t.end();
     });
--- a/test/async-await.js
+++ b/test/async-await.js
@@ -1,0 +1,220 @@
+var tap = require('tap');
+
+var stripFullStack = require('./common').stripFullStack;
+var runProgram = require('./common').runProgram;
+
+var nodeVersion = process.versions.node;
+var majorVersion = nodeVersion.split('.')[0];
+
+if (Number(majorVersion) < 8) {
+    process.exit(0);
+}
+
+tap.test('async1', function (t) {
+    runProgram('async-await', 'async1.js', function (r) {
+        t.same(r.stdout.toString('utf8'), [
+            'TAP version 13',
+            '# async1',
+            'ok 1 before await',
+            'ok 2 after await',
+            '',
+            '1..2',
+            '# tests 2',
+            '# pass  2',
+            '',
+            '# ok'
+        ].join('\n') + '\n\n');
+        t.same(r.exitCode, 0);
+        t.same(r.stderr.toString('utf8'), '');
+        t.end();
+    });
+});
+
+tap.test('async2', function (t) {
+    runProgram('async-await', 'async2.js', function (r) {
+        var stdout = r.stdout.toString('utf8');
+        var lines = stdout.split('\n');
+        lines = lines.filter(function (line) {
+            return ! /^(\s+)at(\s+)<anonymous>$/.test(line);
+        });
+        stdout = lines.join('\n');
+
+        t.same(stripFullStack(stdout), [
+            'TAP version 13',
+            '# async2',
+            'ok 1 before await',
+            'not ok 2 after await',
+            '  ---',
+            '    operator: ok',
+            '    expected: true',
+            '    actual:   false',
+            '    at: Test.myTest ($TEST/async-await/async2.js:$LINE:$COL)',
+            '    stack: |-',
+            '      Error: after await',
+            '          [... stack stripped ...]',
+            '          at Test.myTest ($TEST/async-await/async2.js:$LINE:$COL)',
+            '  ...',
+            '',
+            '1..2',
+            '# tests 2',
+            '# pass  1',
+            '# fail  1'
+        ].join('\n') + '\n\n');
+        t.same(r.exitCode, 1);
+        t.same(r.stderr.toString('utf8'), '');
+        t.end();
+    });
+});
+
+tap.test('async3', function (t) {
+    runProgram('async-await', 'async3.js', function (r) {
+        t.same(r.stdout.toString('utf8'), [
+            'TAP version 13',
+            '# async3',
+            'ok 1 before await',
+            'ok 2 after await',
+            '',
+            '1..2',
+            '# tests 2',
+            '# pass  2',
+            '',
+            '# ok'
+        ].join('\n') + '\n\n');
+        t.same(r.exitCode, 0);
+        t.same(r.stderr.toString('utf8'), '');
+        t.end();
+    });
+});
+
+tap.test('async4', function (t) {
+    runProgram('async-await', 'async4.js', function (r) {
+        t.same(stripFullStack(r.stdout.toString('utf8')), [
+            'TAP version 13',
+            '# async4',
+            'ok 1 before await',
+            'not ok 2 Error: oops',
+            '  ---',
+            '    operator: error',
+            '    expected: |-',
+            '      undefined',
+            '    actual: |-',
+            '      [Error: oops]',
+            '    at: Test.myTest ($TEST/async-await/async4.js:$LINE:$COL)',
+            '    stack: |-',
+            '      Error: oops',
+            '          at Timeout.myTimeout [as _onTimeout] ($TEST/async-await/async4.js:$LINE:$COL)',
+            '          [... stack stripped ...]',
+            '  ...',
+            '',
+            '1..2',
+            '# tests 2',
+            '# pass  1',
+            '# fail  1'
+        ].join('\n') + '\n\n');
+        t.same(r.exitCode, 1);
+        t.same(r.stderr.toString('utf8'), '');
+        t.end();
+    });
+});
+
+tap.test('async5', function (t) {
+    runProgram('async-await', 'async5.js', function (r) {
+        t.same(r.stdout.toString('utf8'), [
+            'TAP version 13',
+            '# async5',
+            'ok 1 before server',
+            'ok 2 after server',
+            'ok 3 before request',
+            'ok 4 after request',
+            'ok 5 should be equal',
+            'ok 6 should be equal',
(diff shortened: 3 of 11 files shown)


Raynos:
This is a much requested feature by my peers @juliangruber @maxharris9 @davidmarkclements

See related issues : 

 - https://github.com/substack/tape/pull/441
 - https://github.com/substack/tape/issues/208
 - https://github.com/substack/tape/issues/358
 - https://github.com/substack/tape/issues/262

ljharb:
> in future versions of node, unhandled rejections may get forwarded to uncaught exceptions.

fwiw this won’t actually happen by default; unhandled rejections are a normal part of the language, and by default, the program needs to continue when one occurs. The node warning message is overly ambitious and thus misleading.

.eslintrc
```diff
@@ -7,4 +7,7 @@
             "named": "never",
         }],
     },
+    "parserOptions": {
+        "ecmaVersion": 2017
```
ljharb: let’s enable this only for the tests that need it; otherwise breaking syntax could be added.

lib/test.js
```diff
@@ -93,7 +93,22 @@ Test.prototype.run = function () {
...
     }
     this.emit('prerun');
-    this._cb(this);
+    var callbackReturn = this._cb(this);
+
+    if (
+        typeof Promise === 'function' &&
+        callbackReturn &&
+        typeof callbackReturn.then === 'function' &&
+        typeof callbackReturn.catch === 'function'
```
ljharb: .catch isn’t required; any thenable should be supported.

lib/test.js
```diff
@@ -93,7 +93,22 @@ Test.prototype.run = function () {
...
+    ) {
+        callbackReturn.catch(function onError(err) {
```
ljharb: since it might not be a real promise, we should wrap it in Promise.resolve first.

lib/test.js
```diff
@@ -93,7 +93,22 @@ Test.prototype.run = function () {
...
+            nextTick(function rethrowError() {
+                throw err
```
ljharb:
```suggestion
                throw err;
```
and throughout

ljharb:
as-is, this is still a breaking change - anyone currently using an async function (or any promise-returning function) with an unhandled rejection will cause an error where none previously existed.

lib/test.js
```diff
@@ -93,7 +93,22 @@ Test.prototype.run = function () {
...
+    var callbackReturn = this._cb(this);
+
+    if (
+        typeof Promise === 'function' &&
+        callbackReturn &&
+        typeof callbackReturn.then === 'function' &&
+        typeof callbackReturn.catch === 'function'
+    ) {
+        callbackReturn.catch(function onError(err) {
+            nextTick(function rethrowError() {
```
ljharb: Im actually confused here - all this seems to be doing is ensuring that the returned promise can’t hook into node’s unhandledRejection event - it’s not actually failing the associated test and allowing tests to continue.

Raynos:
> Im actually confused here - all this seems to be doing is ensuring that the returned promise can’t hook into node’s unhandledRejection event - it’s not actually failing the associated test and allowing tests to continue.

All this does is make exceptions in async functions work. unhandled rejections from the async function passed to `tape(fnName, fn)` will go to uncaught exception and have the default process uncaught exception semantics.

> fwiw this won’t actually happen by default; unhandled rejections are a normal part of the language, and by default, the program needs to continue when one occurs. The node warning message is overly ambitious and thus misleading.

I was at the openjs summit recently and spoke to some node core folks, it seems like we want stricter semantics around unhandled rejections in core, forwarding unhandled rejections to uncaught exceptions might be a change we see in future node versions.

Unhandled rejections are a bug and undefined behavior, they should be handled exactly like uncaught exceptions.

Raynos:
> as-is, this is still a breaking change - anyone currently using an async function (or any promise-returning function) with an unhandled rejection will cause an error where none previously existed.

If you have an async function that throws an exception its "silently" continueing. Your test suite is broken, it's just that the exitCode of the process does not reflect it.

This patch makes the exception you threw visible. 

Technically it is a breaking change, it's a change in semantics. However it is a better user experience for exception handling in async functions and normal functions to be the same in `tape`

.eslintrc
Raynos: I'll see if I can opt in on a per file basis, maybe a nested `.eslintrc` file in one of the test folders is sufficient.

lib/test.js
Raynos: Personally I only care about supporting async functions and not thenable returning functions.

However if you care strongly I can add support for thenable's

lib/test.js
Raynos: I don't understand why that matters in this concrete case. It might be a good practice in general but I think for this patch it's unnecessary.

lib/test.js
Raynos: willfix.

Raynos:
> anyone currently using an async function (or any promise-returning function) with an unhandled rejection will cause an error where none previously existed.

It should be noted that there's very few cases of this happening, if you have a thrown exception in an async function you are probably not calling `t.end()`

The behavior reported in https://github.com/substack/tape/issues/358 is more common for async functions with thrown exceptions, confusion about tests not ending and just hanging.

Tape will report "expected t.end() to be called but test not finished" if you have no handles open in your test process, but most realistic tests will actually leak sockets or timers or something that keeps the event loop open when you throw in an async function. This leads to the output of the process just hanging and tape will not report anything, and the process wont exit because you didnt teardown your http server or whatever.

ljharb:
They’re not a bug, and they’re quite defined - if node exits on them by default, it’s violating the intent of the spec.

ljharb:
I've rebased this, and added a commit that changes the behavior to:
 - fail the test when its callback returns a rejected thenable
 - call `.end()` for you when its callback returns a resolved thenable
 - fail when you've called `.end()` already before returning a resolved thenable (this part would be a breaking change for sure, even if the rest wouldn't be - so i'm open to making this silently work)

Thoughts?

lib/test.js
```diff
@@ -93,7 +93,27 @@ Test.prototype.run = function () {
         this.timeoutAfter(this._timeout);
     }
     this.emit('prerun');
-    this._cb(this);
+    var callbackReturn = this._cb(this);
+
+    if (
+        typeof Promise === 'function' &&
+        callbackReturn &&
+        typeof callbackReturn.then === 'function'
```
juliangruber: why would we only support thenables if promises are available? This seems like it would make it a half-assed thenables implementation, either we support them 100% or not at all?

Raynos:
> fail when you've called .end() already before returning a resolved thenable (this part would be a breaking change for sure, even if the rest wouldn't be - so i'm open to making this silently work)

The existing tests I wrote had `t.end()` in them because that's a pattern people are already doing. The only test in which my patch is changing semantics is the `async-error.js` test.

I think failing on "double end" is too heavy of a breaking change for this patch. It requires fixing tests and code that "just works" before when you upgrade tape.

When we make a breaking change to report the rejected promise error at least you are fixing dubious tests that silently "failed" and are worth looking at and fixing. If you upgrade tape and it makes a breaking change I hope that the break will actually allow you identify tests that silently failed and improved your test suite.

You don't want to upgrade tape and just refactor tests that are "correct"

> call .end() for you when its callback returns a resolved thenable

This is actually a breaking change that will break "correct" tests that have passed before, I'm going to push a test case onto this branch to demonstrate this, but here is an example

```js
tape('my func', async function (t) {
  var res = await fetch('http://localhost:3000')
  var body = await res.json()
  t.equal(body, '"ok"')

  checkDb(function done(data) {
    t.equal(data, 'is good')
    t.end()
  })
})
```

I've seen tests in the wild where people are migrating their code to async/await. They start using for example `node-fetch` and using promises for HTTP but their database driver is still callbacks and they make callback related code in their tests.

This means they are using `t.end()` and if tape were to auto call `t.end()` when the async function completes this test would finish before the database read completed which leads to a `t.equal()` after `t.end()`

I strongly recommend this patch implements support for async functions and makes minimal breaking changes based on how people are using `tape` and async functions in the wild instead of implementing "proper promises" support.

> fail the test when its callback returns a rejected thenable

My original patch has two test cases, `sync-error.js` and `async-error.js` ; I made the smallest change possible to handle the promise returned from an async function such that those two test cases behave the same ( complain on STDERR and exit with code 1 ).

Your change preserves the `exit code 1` semantics but does mean that sync functions and async functions report their failures differently, both of them print the stack to stdout or stderr and both of them fail.

I'm ok with failing the test but I still prefer having identical behavior for sync and async errors. Note that adding a try / catch to make sync exceptions fail the test is probably a bad idea and out of scope for this PR.

Raynos:
CI fails because the anonymous stack frame needs to be stripped, I have another test with an example on how to strip it.

I'm going to push a commit to add another test case and add `t.end()` back into the test cases.

juliangruber:
> This is actually a breaking change that will break "correct" tests that have passed before, I'm going to push a test case onto this branch to demonstrate this, but here is an example
> 
> ```js
> tape('my func', async function (t) {
>   var res = await fetch('http://localhost:3000')
>   var body = await res.json()
>   t.equal(body, '"ok"')
> 
>   checkDb(function done(data) {
>     t.equal(data, 'is good')
>     t.end()
>   })
> })
> ```

To me this should make the test fail as you're mixing control flow primitives. Once the async function completes, the test is done. Anything happening after that is unsafe territory. How would the test harness be able to tell it should wait for an explicit `t.end()` before starting to run the next test?

Raynos:
> How would the test harness be able to tell it should wait for an explicit t.end() before starting to run the next test?

For back compat reasons the test harness cannot implicitly call `t.end()` when the async function completes. You still have to manually call `t.end()` in an `async` function.

juliangruber:
Ok. I'm not really interested in this patch unless it implicitly calls `t.end()` because then I'll need a tape wrapper anyway. tap and ava and mocha all do this, I hope tape can move into the future sooner rather than later...and we know we'll need a breaking change for this anyway, so why not do it right away.

Raynos:
This patch is purely about fixing the exitCode of your test suite when using async functions so that you can trust CI and it doesn't "silently fail"

This patch is not about improved user experience

Accidentally using an `async` function with tape can lead to tests that fail but exit with exitCode 0. 

I'm open to improving the user experience but that should be a seperate PR.

This one is for the bugfix of the exitcode, not for the feature of improved user experience with async functions.

lib/test.js
ljharb: because if they’re not available, we’d have to ship an entire Promises shim to be able to provide spec-compliant behavior.

I think it’s reasonable to gate this functionality behind the almost-certainty that Promise is being shimmed in the final application.

ljharb:
I don’t think we can fix one without providing the other; it’s not currently a bug because tape has no promise support, and a rejected promise is not an error per the intention of the JS spec.

Raynos:
> I don’t think we can fix one without providing the other; it’s not currently a bug because tape has no promise support, and a rejected promise is not an error per the intention of the JS spec.

The user experience for thrown exceptions in async functions is bad. The behavior is different from thrown exceptions in normal functions.

I would recommend you look at this from the perspective of exceptions in functions and not the behavior of promises themselves.

We can add a check to see if the fn is an AsyncFunction if you prefer.

Raynos:
That being said I'm fine with supporting promises if that's what we need to get async functions to work

ljharb:
There’s no way to determine reliably if something is an async function without parsing toString, and an async function is nothing more than a function that returns a Promise - async/await *is* promises, nothing more.

Raynos:
Ok, I think this PR is ready for another review. 

There were a few issues mentioned that are worth discussing

 - [ ] Can we implicitly call `t.end()` when the promise resolves ? This might be a breaking change, see test case that I've added.
 - [ ] Call `t.fail(err)` when the promise rejects or rethrow the error to uncaught exception ? I'm fine with either approach.

r0mflip:
> Can we implicitly call t.end() when the promise resolves ?

Explicit is 👌🏻

> Call t.fail(err) when the promise rejects or rethrow the error to uncaught exception ? I'm fine with either approach.

Fail and call end.

test/async-await.js
```diff
@@ -0,0 +1,247 @@
...
+                ! /Immediate\.next/.test(line);
+        });
+        stderr = lines.join('\n');
+
+        t.same(stderr, '');
+        t.end();
+    });
+});
+
+function runProgram(folderName, fileName, cb) {
```
r0mflip: Move this into `test/common.js`?

lib/test.js
```diff
@@ -93,7 +93,21 @@ Test.prototype.run = function () {
...
+    ) {
+        var self = this;
+        Promise.resolve(callbackReturn).catch(function onError(err) {
```
ljharb: Note: another alternative implementation is:
```jsx
new Promise(function (resolve) {
  resolve(self._cb(self));
}).catch(…);
```
which would ensure that *sync* errors don't kill the whole suite but instead fail the test (which mimics other test runners), and would make sync and async errors behave the same.

test/async-await.js
Raynos: willfix.

lib/test.js
Raynos: This creates zalgo behavior where only exceptions thrown synchronously are caught and exceptions in callbacks go to uncaught handler.

With the `async/await` implementation, if the application developer uses promises and `await` all exceptions, whether synchronous or after `await` are caught in the rejected promise. This only holds true if they don't use any callbacks and rely on promises / `events.once` / `util.promisify`

Raynos:
Addressed feedback & rebased.

lib/test.js
ljharb: It's always the case - whether using async/await or not - that if you use async callbacks and they throw an exception, it won't be caught. That's not creating zalgo behavior, it's just unavoidable.

I don't think it's zalgo that sync exceptions fail the test while non-promise async ones continue to have the same behavior.

Raynos:
> we can, in this PR, or at any time in the future, start catching sync errors in tests and make them fail the tests.

This is something useful to discuss but might be out of scope for this PR.

> It would make total sense to me if it did "if not already ended, end it" 

I'm ok with adding this change. But if the user called `t.end()` we can just silently do nothing since that's already a common pattern of using `async/await` with tape as is.

The pattern of returning from the async function and doing some callback IO without awaiting a promise is less common and maybe worth breaking as an bad practice edge case.

ljharb:
Agreed on number 1 being out of scope - might be worth pursuing later.

Sounds like we have agreement for number 4? I can rebase this PR and add it, and merge. I'll take another pass as well at the other pending breaking changes before cutting a v5.

test/async-await/async5.js
```diff
@@ -0,0 +1,57 @@
...
+            });
+            req.end();
+        })();
+
+        t.ok(true, 'after request');
+
+        res.resume();
+        t.equal(res.statusCode, 200);
+
+        setTimeout(function () {
```
ljharb: note that using `setTimeout` in a world of async/await is a bug. This should be `await`ing a `new Promise`, so that its fate is tied to the overarching promise of the async function.

Raynos: It’s a bad practice but not a bug. 

I’ve seen plenty of people mixing promises, async/await and callbacks

Adding ‘data’ listeners is the biggest example of mixing callbacks and promises.

This is a best practice we will need to educate the tape users on

Raynos:
For the tests :

 - can we add t.end() back into one of the tests to verify it works with and without end
 - can we add back a test with setTimeout and call t.end after async function finishes. It should fail with double end and exitCode 1 I think.

Otherwise lgtm, I might push those tests myself.

test/async-await/async5.js
ljharb: I consider those two things the same thing :-)

If you're returning a promise - which is always the case when using an `async function` - you are explicitly stating that your function resolves to a single future async value. Thus, every single async action you create needs to be tied to the fate of that promise.

If that's not the case, take a callback and don't be an async function/don't return a promise.

I agree educating tape users might be needed.

Raynos: 100% agreed, mixing callbacks & promises is a terrible idea.

But I suspect our users might do it none the less and tape already "supports" async functions if you call `t.end()` and happen to not throw an exception so we might break existing tests

Raynos:
I've added the tests I mentioned I would add.

I think this PR is ready for final review and then I might reason out to substack to see if he want's to lgtm it.

Raynos:
I reached out to substack, I’ll take his silence as trusting us to figure it out.

I’m happy to help prep a branch for release, since this is semver major maybe we want other patches to go out as well, cc @ljharb ?

ljharb:
Let me double check to make sure all the non-major stuff that's ready is merged and released; and then we can just start bringing in semver-major stuff into master, including this one (i'll do a final rereview of this one as well first)

Raynos:
@ljharb would you like some help rolling up a new major release ?

ljharb:
Thanks, but i’m not quite ready yet :-) im making some changes in deep-equal first, so i can roll that into the major. Sometime next week i hope to have it all prepared.

Raynos:
Sounds great, thanks for keeping me posted :)

ljharb:
ftr: I'm waiting on a particular user to maybe hand over 4 package names; which I'd use to make robust brand checking methods to help add proper collection support to deep-equal, at which point i'd release a v2 of it, and then merge this along with updating that. Hopefully next week? I haven't forgotten!

Raynos:
@ljharb friendly reminder :D

ljharb:
Thanks, I haven't forgotten. I'm waiting on a few package names; once they're transferred I'll publish them, bump deep-equal's major, and then merge this and bump tape's major.

ljharb:
I'll get all the pending breaking changes in, and publish a prerelease for v5 soon.

Raynos:
Nice :) I will update a bunch of my apps & libraries to the pre-release once available.

Raynos:
I ran into the unhandled rejection issue with tape again.

I thought I fixed this almost a year ago :disappointed: 

Can we release v5 officially as `latests` ?

ljharb:
`tape@next` has been available for awhile; I've run into a number of bugs, which have delayed the regular v5 from being released as I increased usage.

I'm now using `v5.0.0-next.5` on 200+ of my projects, so I'm pretty confident that it's ready to go. I'll plan to release v5 later today or tomorrow.

Raynos:
Awesome, thanks! I appreciate it.

ljharb:
v5.0.0 is released.

Raynos:
Nice I released `@pre-bundled/tape` v5.0.0 ( https://www.npmjs.com/package/@pre-bundled/tape/v/5.0.0 ). 

:tada:

3cp:
Love the new feature! Should tape two more assertions to handle async throw? Like tape-promise does:

https://github.com/jprichardson/tape-promise#new-assertions

ljharb:
@3cp a) that seems like it would conflict with "the test ends when the returned-to-tape promise is  settled", b) let's discuss new features in new issues :-)

PR #473, opened 2019-06-22
[Fix] emit skipped tests as objects

r0mflip:
When using `objectMode` skipped tests are emitted as comments, they are actual tests that need to be emitted as objects.

Test code:
``` js
const test = require('tape');
const printer = new require('stream').Writable({objectMode: true});
printer.write = printer.end = obj => {
  console.log(obj);
};

test.createStream({objectMode: true})
.pipe(printer);

test('parent', t => {
  t.equal(true, true);
  t.test('child1', {skip: 'not installed'}, t2 => {
    t2.equal(true, true);
    t2.equal(true, false);
    t2.end();
  });
  t.test('child2', {todo: 'will get done soon'}, t3 => {
    t3.equal(true, false);
    t3.equal(true, true);
    t3.end();
  });
  t.test('child3', {todo: 'will do'});
  t.equal(true, true);
  t.equal(true, true);
  t.end();
});
```

When used normally:
```
{ type: 'test', name: 'parent', id: 0 }
{
  id: 0,
  ok: true,
  skip: undefined,
  todo: false,
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 0,
  type: 'assert'
}
{
  id: 1,
  ok: true,
  skip: undefined,
  todo: false,
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 0,
  type: 'assert'
}
{
  id: 2,
  ok: true,
  skip: undefined,
  todo: false,
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 0,
  type: 'assert'
}
SKIP child1                                                 <- plain text
{ type: 'end', test: 1 }                                    <- no object with testId 1
{ type: 'test', name: 'child2', id: 2, parent: 0 }
{
  id: 0,
  ok: false,
  skip: undefined,
  todo: 'will get done soon',
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: false,
  functionName: 'Test.<anonymous>',
  file: 'D:\\dev\\testing\\tape-test\\test.js:19:8',
  line: 19,
  column: 8,
  at: 'Test.<anonymous> (D:\\dev\\testing\\tape-test\\test.js:19:8)',
  test: 2,
  type: 'assert'
}
{
  id: 1,
  ok: true,
  skip: undefined,
  todo: 'will get done soon',
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 2,
  type: 'assert'
}
{ type: 'end', test: 2 }
{ type: 'end', test: 3 }                                       <- no object with testId 3
{ type: 'end', test: 0 }
```

After patch:
```
{ type: 'test', name: 'parent', id: 0, skip: false, todo: false }
{
  id: 0,
  ok: true,
  skip: undefined,
  todo: false,
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 0,
  type: 'assert'
}
{
  id: 1,
  ok: true,
  skip: undefined,
  todo: false,
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 0,
  type: 'assert'
}
{
  id: 2,
  ok: true,
  skip: undefined,
  todo: false,
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 0,
  type: 'assert'
}
{
  type: 'test',
  name: 'child1',
  id: 1,                                                         <- object with testId 1
  skip: 'not installed',
  todo: false,
  parent: 0
}
{ type: 'end', test: 1 }                                         <- reference exists
{
  type: 'test',
  name: 'child2',
  id: 2,
  skip: false,
  todo: 'will get done soon',
  parent: 0
}
{
  id: 0,
  ok: false,
  skip: undefined,
  todo: 'will get done soon',
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: false,
  functionName: 'Test.<anonymous>',
  file: 'D:\\dev\\testing\\tape-test\\test.js:19:8',
  line: 19,
  column: 8,
  at: 'Test.<anonymous> (D:\\dev\\testing\\tape-test\\test.js:19:8)',
  test: 2,
  type: 'assert'
}
{
  id: 1,
  ok: true,
  skip: undefined,
  todo: 'will get done soon',
  name: 'should be equal',
  operator: 'equal',
  objectPrintDepth: 5,
  actual: true,
  expected: true,
  test: 2,
  type: 'assert'
}
{ type: 'end', test: 2 }
{
  type: 'test',
  name: 'child3',
  id: 3,                                                         <- object with testId 3
  skip: false,
  todo: 'will do',
  parent: 0
}
{ type: 'end', test: 3 }                                         <- reference exists
{ type: 'end', test: 0 }
```

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -41,7 +41,9 @@ Results.prototype.createStream = function (opts) {
                 var row = {
                     type: 'test',
                     name: t.name,
-                    id: id
+                    id: id,
+                    skip: t._skip,
+                    todo: t._todo
                 };
                 if (has(extra, 'parent')) {
                     row.parent = extra.parent;
@@ -97,7 +99,10 @@ Results.prototype._watch = function (t) {
     var self = this;
     var write = function (s) { self._stream.queue(s); };
     t.once('prerun', function () {
-        write('# ' + t.name + '\n');
+        var premsg = '';
+        if (t._skip) premsg = 'SKIP ';
+        else if (t._todo) premsg = 'TODO ';
+        write('# ' + premsg + t.name + '\n');
     });
 
     t.on('result', function (res) {
--- a/test/objectMode.js
+++ b/test/objectMode.js
@@ -1,0 +1,70 @@
+var tap = require('tap');
+var tape = require('../');
+var forEach = require('for-each');
+var through = require('through');
+
+tap.test('object results', function (assert) {
+    var printer = through({ objectMode: true });
+    var objects = [];
+
+    printer.write = function (obj) {
+        objects.push(obj);
+    };
+
+    printer.end = function (obj) {
+        if (obj) objects.push(obj);
+
+        var todos = 0;
+        var skips = 0;
+        var testIds = [];
+        var endIds = [];
+        var asserts = 0;
+
+        assert.equal(objects.length, 13);
+
+        forEach(objects, function (obj) {
+            if (obj.type === 'assert') {
+                asserts++;
+            } else if (obj.type === 'test') {
+                testIds.push(obj.id);
+
+                if (obj.skip) {
+                    skips++;
+                } else if (obj.todo) {
+                    todos++;
+                }
+            } else if (obj.type === 'end') {
+                endIds.push(obj.text);
+                // test object should exist
+                assert.notEqual(testIds.indexOf(obj.test), -1);
+            }
+        });
+
+        assert.equal(asserts, 5);
+        assert.equal(skips, 1);
+        assert.equal(todos, 2);
+        assert.equal(testIds.length, endIds.length);
+        assert.end();
+    };
+
+    tape.createStream({ objectMode: true })
+        .pipe(printer);
+
+    tape('parent', function (t1) {
+        t1.equal(true, true);
+        t1.test('child1', {skip: true}, function (t2) {
+            t2.equal(true, true);
+            t2.equal(true, false);
+            t2.end();
+        });
+        t1.test('child2', {todo: true}, function (t3) {
+            t3.equal(true, false);
+            t3.equal(true, true);
+            t3.end();
+        });
+        t1.test('child3', {todo: true});
+        t1.equal(true, true);
+        t1.equal(true, true);
+        t1.end();
+    });
+});
--- a/test/todo.js
+++ b/test/todo.js
@@ -15,7 +15,7 @@ tap.test('tape todo test', function (assert) {
             'TAP version 13\n'
             + '# success\n'
             + 'ok 1 this test runs\n'
-            + '# failure\n'
+            + '# TODO failure\n'
             + 'not ok 2 should never happen # TODO\n'
             + '  ---\n'
             + '    operator: fail\n'
--- a/lib/test.js
+++ b/lib/test.js
@@ -83,16 +83,13 @@ function Test(name_, opts_, cb_) {
 }
 
 Test.prototype.run = function () {
-    if (this._skip) {
-        this.comment('SKIP ' + this.name);
-    }
+    this.emit('prerun');
     if (!this._cb || this._skip) {
         return this._end();
     }
     if (this._timeout != null) {
         this.timeoutAfter(this._timeout);
     }
-    this.emit('prerun');
     this._cb(this);
     this.emit('run');
 };
--- a/test/exit.js
+++ b/test/exit.js
@@ -155,7 +155,7 @@ tap.test('todo passing', function (t) {
     var tc = function (rows) {
         t.same(stripFullStack(rows.toString('utf8')), [
             'TAP version 13',
-            '# todo pass',
+            '# TODO todo pass',
             'ok 1 should be truthy # TODO',
             '',
             '1..1',
@@ -179,7 +179,7 @@ tap.test('todo failing', function (t) {
     var tc = function (rows) {
         t.same(stripFullStack(rows.toString('utf8')), [
             'TAP version 13',
-            '# todo fail',
+            '# TODO todo fail',
             'not ok 1 should be truthy # TODO',
             '  ---',
             '    operator: ok',
--- a/test/todo_single.js
+++ b/test/todo_single.js
@@ -13,7 +13,7 @@ tap.test('tape todo test', function (assert) {
         assert.equal(
             stripFullStack(body.toString('utf8')),
             'TAP version 13\n'
-            + '# failure\n'
+            + '# TODO failure\n'
             + 'not ok 1 should be equal # TODO\n'
             + '  ---\n'
             + '    operator: equal\n'


lib/results.js
```diff
@@ -139,8 +144,8 @@ function encodeResult(res, count) {
     output += (res.ok ? 'ok ' : 'not ok ') + count;
     output += res.name ? ' ' + res.name.toString().replace(/\s+/g, ' ') : '';
 
-    if (res.skip) output += ' # SKIP';
-    else if (res.todo) output += ' # TODO';
+    if (res.skip) output += ' # SKIP' + (typeof res.skip === 'string' ? ' ' + res.skip : '');
```
r0mflip: Sorry this change crept in, It is unintentional regarding this PR. Sorry.

lib/results.js
```diff
@@ -97,7 +99,10 @@ Results.prototype._watch = function (t) {
     var self = this;
     var write = function (s) { self._stream.queue(s) };
     t.once('prerun', function () {
-        write('# ' + t.name + '\n');
+        var premsg = '';
+        if (t._todo) premsg = 'TODO ';
```
r0mflip: This is done for consistency (respective tests are updated)

lib/results.js
```diff
@@ -97,7 +99,10 @@ Results.prototype._watch = function (t) {
...
+        if (t._skip) premsg = 'SKIP ';
```
ljharb: can a test be both todo and skip?

lib/results.js
ljharb: let’s revert it then?

lib/results.js
```diff
@@ -139,8 +144,8 @@ function encodeResult(res, count) {
     output += (res.ok ? 'ok ' : 'not ok ') + count;
     output += res.name ? ' ' + res.name.toString().replace(/\s+/g, ' ') : '';
 
-    if (res.skip) output += ' # SKIP';
-    else if (res.todo) output += ' # TODO';
+    if (res.skip) output += ' # SKIP' + (typeof res.skip === 'string' ? ' ' + res.skip : '');
+    else if (res.todo) output += ' # TODO' + (typeof res.todo === 'string' ? ' ' + res.todo : '');
```
ljharb: same here?

test/todo.js
```diff
@@ -15,7 +15,7 @@ tap.test('tape todo test', function (assert) {
             'TAP version 13\n'
             + '# success\n'
             + 'ok 1 this test runs\n'
-            + '# failure\n'
+            + '# TODO failure\n'
```
ljharb: hm, there’s a suffix on all the assertions, why would the description get a prefix?

r0mflip: This was to maintain consistency with `# SKIP something`

lib/results.js
r0mflip: This better suits with the TODO descriptions in https://testanything.org/tap-version-13-specification.html#todo-tests

It _maybe_ useful, these messages also get passed in the JSON results which can help in diagnostics on the consumer side.

ljharb: ah - so this adds the assertion message after the ‘’TODO” marker? some tests would help illustrate this :-)

lib/results.js
r0mflip: Here `skip` takes precidence https://github.com/substack/tape/blob/db1aa0682204265f9765c7ddea931726e01da8b8/lib/test.js#L87-L89 so I thought we might not need an else. When the user passes in both `todo` and `skip` in `opts`, skip takes precidence.

lib/results.js
r0mflip: https://github.com/substack/tape/pull/473#discussion_r296451255

r0mflip:
> Please add tests covering your new changes.

Will add tests for `objectMode` stream and stuff, if all regarding changes in this PR are ok

lib/results.js
ljharb: It seems useful to keep the else here, just in case someone calls into this directly.

ljharb:
Please do; tests help review of code.

lib/results.js
r0mflip: Will do

lib/results.js
r0mflip: https://github.com/substack/tape/pull/473#discussion_r296451500

r0mflip:
Node 0.8 has a weird issue that I don't understand why only it fails.

lib/results.js
ljharb: did this not require any test changes? if not, can we add tests that illustrate this change?

test/todo_explanation.js
```diff
@@ -0,0 +1,58 @@
...
+        assert.equal(
+            stripFullStack(body.toString('utf8')),
+            'TAP version 13\n'
+            + '# success\n'
+            + 'ok 1 this test runs\n'
+            + '# TODO incomplete test1\n'
+            + 'not ok 2 needs insight # TODO\n'
+            + '  ---\n'
+            + '    operator: fail\n'
+            + '    at: Test.<anonymous> ($TEST/todo_explanation.js:$LINE:$COL)\n'
```
ljharb: the test on 0.8 seems to want `Test.tap.test.test.todo` here instead of `Test.<anonymous>` :-/

I'm confused because we have many other tests using `Test.<anonymous>` that don't have special handling for node 0.8.

test/todo_explanation.js
```diff
@@ -0,0 +1,58 @@
...
+            + '  ...\n'
+            + '# TODO incomplete test2\n'
+            + 'not ok 3 check description # TODO incomplete sentiment\n'
+            + '  ---\n'
+            + '    operator: fail\n'
+            + '    at: Test.<anonymous> ($TEST/todo_explanation.js:$LINE:$COL)\n'
```
ljharb: also here

test/todo_explanation.js
r0mflip: I get that, but only why node 0.8 😭 I've gone through the source and debugged it a thousand times but only node 0.8 failes for only `todo_explanation.js`. It's just a copy of `todo.js`

ljharb: I'm wondering if the change to `emit('prerun') somehow changed things for the v8 version 0.8 is using.

We could update `test.common`, perhaps, for node 0.8 only, to swap out `Test.tap.test.test.todo` for `Test.<anonymous>`?

lib/results.js
r0mflip: Adding these tests I found a new problem:
``` js
tape('incomplete test with explanation', function (t) {
  // var platform = process.platform; // something like this needed
  var platform = 'win32';

  t.fail('run sh', { skip: platform === 'win32' });
  t.fail('run openssl', { skip: platform === 'win32' && 'can\'t run on windows platforms' });
  t.pass('this runs');
  t.end();
});
```

Even though `skip` is passed in `extra` the tests fail with error, IMO this needs another fix (and tests).

Probably these lines here
https://github.com/substack/tape/blob/c6c4aceccf52fb189732b95fd049b4e110d7868f/lib/test.js#L223-L227

will need to be changed to
``` js
ok = ok || res.skip; // <- goes before line 223
```

ljharb: lol, i wouldn't even expect `skip` to work in an assertion, only in a `test` or `t.test` call.

r0mflip: Well this line is related to any `t.assert`, so 🤷🏻‍♂️

r0mflip: > expect skip to work

To be used?

test/todo_explanation.js
r0mflip: > I'm wondering if the change to `emit('prerun') somehow changed things for the v8 version 0.8 is using.

I tried by using the latest commit from `master`, It still has the same problem.

My [naive] understanding is that in V8 version of 0.8, the `stack` part assumes the `{todo: 'something'}` object as one of the stack layer. The same happens for my local `skip_explanation.js` test. The test in https://github.com/substack/tape/pull/473#discussion_r296560029 comment has the function in stack pointing to `Test.tap.test.platform`.

r0mflip:
Removed loose patches. Current changes include:
1. Emit all objects (tests added)
3. Add prefix (`# TODO ...`, `# SKIP ...`) to comments (tests altered accordingly)

Will try to add descriptive messages in another PR

lib/results.js
ljharb: I’ve never passed options into an assertion - I’m surprised it does anything.

r0mflip: It does actually work, and as per TAP individual explanations need this. Anyways I've removed this from current branch. I will perhaps try to make another PR for it and we can discuss it their.

I've added my final tests and patches. This PR is done for, please review.

test/objectMode.js
```diff
@@ -0,0 +1,70 @@
+var tap = require('tap');
+var tape = require('../');
+var forEach = require('for-each');
+var through = require('through');
+
+tap.test('object results', function (assert) {
+    var printer = through({ objectMode: true });
```
ljharb, line 7: apparently, the `through` package has never supported an `objectMode` option, so i'm not sure what this is testing.

Raynos: objectMode does not exist as an option here, but the implementation is so trivial that it works with strings, objects & buffers by default. Where as node streams require opting in to allow objects with an objectMode flag.

PR #476, opened 2019-06-26
[Feat] Add descriptive messages for skipped asserts

r0mflip:
Now asserts can take `skip` as a `string` or `boolean`, strings get printed as the reason for skipping the assert.

```
(ok|not ok) 3 <assert message> # SKIP <reason for skipping>
```

### DIFF
--- a/lib/results.js
+++ b/lib/results.js
@@ -16,6 +16,10 @@ var nextTick = typeof setImmediate !== 'undefined'
 module.exports = Results;
 inherits(Results, EventEmitter);
 
+function coalesceWhiteSpaces(str) {
+    return String(str).replace(/\s+/g, ' ');
+}
+
 function Results() {
     if (!(this instanceof Results)) return new Results;
     this.count = 0;
@@ -102,7 +106,7 @@ Results.prototype._watch = function (t) {
         var premsg = '';
         if (t._skip) premsg = 'SKIP ';
         else if (t._todo) premsg = 'TODO ';
-        write('# ' + premsg + t.name + '\n');
+        write('# ' + premsg + coalesceWhiteSpaces(t.name) + '\n');
     });
 
     t.on('result', function (res) {
@@ -142,10 +146,13 @@ Results.prototype.close = function () {
 function encodeResult(res, count) {
     var output = '';
     output += (res.ok ? 'ok ' : 'not ok ') + count;
-    output += res.name ? ' ' + res.name.toString().replace(/\s+/g, ' ') : '';
+    output += res.name ? ' ' + coalesceWhiteSpaces(res.name) : '';
 
-    if (res.skip) output += ' # SKIP';
-    else if (res.todo) output += ' # TODO';
+    if (res.skip) {
+        output += ' # SKIP' + ((typeof res.skip === 'string') ? ' ' + coalesceWhiteSpaces(res.skip) : '');
+    } else if (res.todo) {
+        output += ' # TODO' + ((typeof res.todo === 'string') ? ' ' + coalesceWhiteSpaces(res.todo) : '');
+    };
 
     output += '\n';
     if (res.ok) return output;
--- a/test/todo_explanation.js
+++ b/test/todo_explanation.js
@@ -1,0 +1,70 @@
+var tap = require('tap');
+var tape = require('../');
+var concat = require('concat-stream');
+
+var common = require('./common');
+var stripFullStack = common.stripFullStack;
+
+tap.test('tape todo test', { todo: process.versions.node.match(/0\.8\.\d+/) ? 'Fails on node 0.8': false }, function (assert) {
+    var test = tape.createHarness({ exit: false });
+    assert.plan(1);
+
+    test.createStream().pipe(concat(function (body) {
+        assert.equal(
+            stripFullStack(body.toString('utf8')), [
+                'TAP version 13',
+                '# success',
+                'ok 1 this test runs',
+                '# TODO incomplete test1',
+                'not ok 2 check output # TODO',
+                '  ---',
+                '    operator: equal',
+                '    expected: false',
+                '    actual:   true',
+                '    at: Test.<anonymous> ($TEST/todo_explanation.js:$LINE:$COL)',
+                '  ...',
+                'not ok 3 check vars output # TODO name conflict',
+                '  ---',
+                '    operator: equal',
+                '    expected: 0',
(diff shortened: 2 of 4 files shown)


ljharb:
Is this a convenience feature, or something that you'd expect as part of any tap client? Are their other TAP-producing runners that do this?

r0mflip:
TAP indicates that any individual assert can be skipped or marked todo. `node-tap` allows this and this also plays really well with `tap-parser`. The same should work for todo, but node 0.8 has problems with the stack.

``` js
test('this skipps', { skip: true }, function (t) {
    t.fail('doesn\'t run');
    t.fail('this doesn\'t run too', { skip: false });
    t.end();
});

test('some tests might skip', function (t) {
    t.pass('this runs');
    t.fail('failing assert is skipped', { skip: true });
    t.pass('this runs');
    t.end();
});

test('incomplete test', function (t) {
    // var platform = process.platform; something like this needed
    var platform = 'win32';

    t.pass('run sh', { skip: platform !== 'win32' });
    t.pass('run openssl', { skip: platform === 'win32' });
    t.end();
});

test('incomplete test with explanation', function (t) {
    // var platform = process.platform; something like this needed
    var platform = 'win32';

    t.fail('run sh (conditional skip)', { skip: platform === 'win32' });
    t.fail('run openssl', { skip: platform === 'win32' && 'can\'t run on windows platforms' });
    t.pass('this runs');
    t.end();
});
```

`tap-parser` output
``` js
[
  [ 'version', 13 ],
  [ 'comment', '# SKIP this skipps\n' ],
  [ 'comment', '# some tests might skip\n' ],
  [
    'assert',
    Result { ok: true, id: 1, name: 'this runs', fullname: '' }
  ],
  [
    'assert',
    Result {
      ok: true,
      id: 2,
      skip: true,
      name: 'failing assert is skipped',
      fullname: ''
    }
  ],
  [
    'assert',
    Result { ok: true, id: 3, name: 'this runs', fullname: '' }
  ],
  [ 'comment', '# incomplete test\n' ],
  [
    'assert',
    Result { ok: true, id: 4, name: 'run sh', fullname: '' }
  ],
  [
    'assert',
    Result {
      ok: true,
      id: 5,
      skip: true,
      name: 'run openssl',
      fullname: ''
    }
  ],
  [ 'comment', '# incomplete test with explanation\n' ],
  [
    'assert',
    Result {
      ok: true,
      id: 6,
      skip: true,
      name: 'run sh (conditional skip)',
      fullname: ''
    }
  ],
  [
    'assert',
    Result {
      ok: true,
      id: 7,
      skip: "can't run on windows platforms",
      name: 'run openssl',
      fullname: ''
    }
  ],
  [
    'assert',
    Result { ok: true, id: 8, name: 'this runs', fullname: '' }
  ],
  [ 'plan', { start: 1, end: 8 } ],
  [ 'comment', '# tests 8\n' ],
  [ 'comment', '# pass  8\n' ],
  [ 'comment', '# ok\n' ],
  [ 'comment', '# skip: 4\n' ],
  [
    'complete',
    FinalResults {
      ok: true,
      count: 8,
      pass: 8,
      fail: 0,
      bailout: false,
      todo: 0,
      skip: 4,
      plan: FinalPlan {
        start: 1,
        end: 8,
        skipAll: false,
        skipReason: '',
        comment: ''
      },
      failures: []
    }
  ]
]
```

lib/results.js
```diff
@@ -144,7 +144,7 @@ function encodeResult(res, count) {
     output += (res.ok ? 'ok ' : 'not ok ') + count;
     output += res.name ? ' ' + res.name.toString().replace(/\s+/g, ' ') : '';
 
-    if (res.skip) output += ' # SKIP';
+    if (res.skip) output += ' # SKIP' + ((typeof res.skip === 'string') ? ' ' + res.skip : '');
```
ljharb: what happens if the message contains a newline?

Let's add some test coverage for that.

lib/results.js
```diff
@@ -142,10 +146,13 @@ Results.prototype.close = function () {
...
     output += (res.ok ? 'ok ' : 'not ok ') + count;
-    output += res.name ? ' ' + res.name.toString().replace(/\s+/g, ' ') : '';
+    output += res.name ? ' ' + coalesceWhiteSpaces(res.name) : '';
 
-    if (res.skip) output += ' # SKIP' + ((typeof res.skip === 'string') ? ' ' + res.skip.toString().replace(/\s+/g, ' ') : '');
-    else if (res.todo) output += ' # TODO';
+    if (res.skip) {
+        output += ' # SKIP' + ((typeof res.skip === 'string') ? ' ' + coalesceWhiteSpaces(res.skip) : '');
+    } else if (res.todo) {
+        output += ' # TODO';
```
r0mflip: We could do the same thing for TODO too, but I don't know how to fix the test for node 0.8
``` js
output += ' # TODO' + ((typeof res.todo === 'string') ? ' ' + coalesceWhiteSpaces(res.todo) : '');
```

lib/results.js
```diff
@@ -16,6 +16,10 @@ var nextTick = typeof setImmediate !== 'undefined'
 module.exports = Results;
 inherits(Results, EventEmitter);
 
+function coalesceWhiteSpaces(str) {
+    return str.toString().replace(/\s+/g, ' ');
```
ljharb:
```suggestion
    return String(str).replace(/\s+/g, ' ');
```
also better than .toString :-)

any reason not to only replace newlines?

lib/results.js
ljharb: let's do it, and i can take a look

lib/results.js
r0mflip: > any reason not to only replace newlines?

I just copied it from https://github.com/substack/tape/blob/master/lib/results.js#L145 It felt like a sane thing to do.

test/todo_explanation.js
```diff
@@ -0,0 +1,70 @@
+var tap = require('tap');
+var tape = require('../');
+var concat = require('concat-stream');
+
+var common = require('./common');
+var stripFullStack = common.stripFullStack;
+
+tap.test('tape todo test', { todo: process.versions.node.match(/0\.8\.\d+/) ? 'Fails on node 0.8': false }, function (assert) {
```
r0mflip: I've temporarily marked it as `todo`

lib/results.js
ljharb: Fair enough!

PR #481, opened 2019-08-06
Browserify build

arindam1993:
`tape` currently works in browsers, but does not work in browsers OOTB and requires significant build configuration to handle node builtins and globals on the library consumer side.

This PR adds a browserify build to the repo, and points the `browser` entry in `package.json` to it.
This a similar approach to that taken by threejs of storing build output in the the repo.
https://github.com/mrdoob/three.js/tree/dev/build

Skimming through the issues, it seems like `tape` devs have been hesitant to handle this and I understand why.

Another approach I can think about is publishing having the `dist/` folder be a separate  subpackage, and using `npm publish` within the subpage on CI to push an updated bundle to npm whenever a new version of tape is released.

### DIFF
--- a/build-browser.js
+++ b/build-browser.js
@@ -21,8 +21,7 @@ modifyPackageJson(function(packageJson) {
     return clone;
 });
 
-var b = browserify()
-    .add(BUNDLE_INPUT)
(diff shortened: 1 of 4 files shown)


ljharb:
Closing for now; happy to reopen if there's a path forward here.

PR #487, opened 2019-12-10
add: link to replit run

eankeen:
i think it would be pretty cool to let people see how this code works (and edit / preview the program) right in their browser, without any manual downloads or installation. i'm thinking we can add a 'run on repl.it' button, which redirects to something like [this](https://repl.it/@eankeen/run-tape):

image

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,2 +1,2 @@
 language = "nodejs"
-run = "yarn install && cd example && node array.js"
+run = "npm install && cd example && node array.js"
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "nodejs"
+run = "yarn install && cd example && node array.js"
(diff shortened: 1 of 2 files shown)


ljharb:
We'd not want to link to something that uses an unofficial package installer, so this is a great idea if you can use npm instead of yarn (ie, if you can make the repl.it actually match this repo, which uses npm)

eankeen:
oh gotchia - yes i can match the commands

.replit
```diff
@@ -0,0 +1,2 @@
+language = "nodejs"
+run = "npm install && cd example && node array.js"
```
ljharb:
```suggestion
run = "npm install && cd example && node array"
```
altho i wonder if we should make an `index.js` that requires all the other files, so we can run all the examples here?

ljharb:
@eankeen did you delete your fork? :-( if so, could you try recreating it, so I can land this?

ljharb:
@eankeen I'd still like to land this, but it appears your fork has been deleted.

PR #489, opened 2019-12-30
[Breaking] fail any assertion after `.end()` is called

### LINKED ISSUE #264: t.end not honored
``` js
var test = require('tape');

test('something', t => {
    t.end();
    t.ok('bananas');
    t.pass('pass');
});
```

==>

```
cdieringer@Snapper-osx:~/node/test-tape-end$ node index.js
TAP version 13
# something
ok 1 (unnamed assert)
ok 2 pass

1..2
# tests 2
# pass  2

# ok
```

this is unexpected.  i would expect t.end not not permit subsequent tests, or otherwise error on them.

ljharb:
Fixes #264.

The intentionally-failing `async5` test could be made to pass (both before and after this change) by wrapping the final setTimeout in a `new Promise` that's returned or awaited.

Before this change, it could be weirdly made to pass by removing the `t.end()` call; after this change, all assertions after the test is complete would have to be removed (barring the proper fix mentioned above).


ljharb:
cc @cdaringe

PR #490, opened 2019-12-30
Avoid setting message property on primitives; use strict to catch this

aredridel:


ljharb:
Thanks; can you add a regression test that would have failed?

ljharb:
Specifically, this doesn't seem like a bug, just a refactor, since in sloppy mode mutating a non-nullish primitive is a silent failure?

aredridel:
Yeah, it only happens if this code is loaded in strict mode, which loading as a module implies. I found it because I was trying to convert the code to ESM.

ljharb:
Gotcha, thanks for clarifying.

Code that's authored in sloppy mode is not safe to run in strict mode in a generic sense; as for converting to ESM, node's ESM integration means that a simple ESM wrapper that imports the CJS is the most compatible way to go - what's the benefit of converting it to ESM fully?

lib/test.js
```diff
@@ -537,7 +538,7 @@ Test.prototype['throws'] = function (fn, expected, msg, extra) {
         fn();
     } catch (err) {
         caught = { error: err };
-        if (err != null && (!isEnumerable(err, 'message') || !has(err, 'message'))) {
+        if (typeof err === 'object' && err != null && (!isEnumerable(err, 'message') || !has(err, 'message'))) {
```
ljharb: This won't let a function through, altho it'd be weird to throw a function. i'll tweak the check a bit.

aredridel:
In my case, just trying to figure out what a truly modern test runner would look like. Tape is nice in that it's fully featured but small enough to approach.

PR #492, opened 2020-01-07
feat: Add -i flag to specify ignore files

### LINKED ISSUE #428: Ability to ignore node_modules or files in .gitignore by default

r0mflip:
Fixes #428 

Confirming to https://github.com/substack/tape/issues/428#issuecomment-367939098

Closes #451.

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -4,13 +4,15 @@
 
 var resolveModule = require('resolve').sync;
 var resolvePath = require('path').resolve;
+var readFileSync = require('fs').readFileSync;
 var parseOpts = require('minimist');
 var glob = require('glob');
+var ignore = require('dotignore');
 
 var opts = parseOpts(process.argv.slice(2), {
-    alias: { r: 'require' },
-    string: 'require',
-    default: { r: [] }
+    alias: { r: 'require', i: 'ignore' },
+    string: ['require', 'ignore'],
+    default: { r: [], i: null }
 });
 
 var cwd = process.cwd();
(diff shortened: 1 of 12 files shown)


bin/tape
```diff
@@ -29,6 +31,15 @@ opts.require.forEach(function (module) {
     }
 });
 
+// Load .gitignore from cwd
+var ignoreStr = '';
+try {
+    ignoreStr = readFileSync(resolvePath(cwd, opts.ignore === '' ? '.gitignore' : opts.ignore), 'UTF-8');
```
ljharb: nit:
```suggestion
    ignoreStr = readFileSync(resolvePath(cwd, opts.ignore === '' ? '.gitignore' : opts.ignore), 'utf-8');
```

bin/tape
```diff
@@ -29,6 +31,15 @@ opts.require.forEach(function (module) {
...
+} catch (e) {}
```
ljharb: i think this should probably exit with an error if `opts.ignore` is provided and the file read doesn't work?

bin/tape
```diff
@@ -39,6 +50,10 @@ opts._.forEach(function (arg) {
     }
 
     files.forEach(function (file) {
+        if (matcher.shouldIgnore(file)) {
```
ljharb: instead of this, let's `files.filter(function (file) { return !matcher.shouldIgnore(file); }).forEach`?

r0mflip:
Sorry I didn't look at any of your comments and review. I am on the other side of the world as you. Timing issues 😝

substack:
This is a really weird feature. tape assumes git now and reads gitignore files? I am very against this.

ljharb:
@substack are you only against the default of `.gitignore`, or against the entire `-i` feature?

PR #496, opened 2020-02-25
Add link to tape-player (in-process reporter) to readme.

DavidAnson:


DavidAnson:
If you do not approve of this approach, I will not be offended. Just reject the PR.

ljharb:
Seems like it might be easier to make a test runner (a binary) instead of a side-effecting require.

DavidAnson:
I dunno, this was pretty easy. :)

But seriously, I’m not sure I understand the suggestion. If you don’t mind explaining the proposal a bit more, I’d appreciate it. I’m happy to use `tape`, I just didn’t want to pipe to a formatter.

ljharb:
That's what I mean - make a new runner, so instead of `tape foo`, which they'd expect to pipe to a TAP formatter, `tape-player foo`, which just prints it out formatted.

DavidAnson:
I get it, thanks, but it seems weirder to me than this and maybe trickier? With how I’ve done it, the use of `tape` is clear in the script command and the behavior of the exit code obvious. Again, no hard feelings if you reject this PR. This works for me, but I may be the only one who likes it.

ljharb:
I think it's fine, but let's move it under "other" instead. It's not a formatter, because it's not configurable or swappable - it's just something that patches tape to print "not TAP".

PR #498, opened 2020-03-02
When t.end() is not called, print the name.

Raynos:
If you forget to call `t.end()` on some test somewhere
and all the other tests pass then you will get a warning
about not having called t.end().

This warning misses which tests should have ended.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -214,7 +214,7 @@ Test.prototype._exit = function () {
             exiting: true
         });
     } else if (!this.ended) {
-        this.fail('test exited without ending', {
+        this.fail('test exited without ending: ' + this.name, {
             exiting: true
         });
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -4,4 +4,5 @@
 
 # Only apps should have lockfiles
 yarn.lock
+.vscode/
```
ljharb: IDE-specific files should go in one's global gitignore, not per project

ljharb:
I'll have a v4.x and a v5 prerelease out soon that contains this PR.

.gitignore
Raynos: Oh that's a great point, I don't know if i have a `~/.gitignore` in my dotfiles

PR #505, opened 2020-04-26
upgrade the dependencies to their latest version

osdevisnot:
This pr upgrades `tap` and `tap-parser` to their latest version. I also tool the opportunity to upgrade `concat-stream`.

Looking forward to any feedback.

Few notable upstream changes:
- `tap-parser` changed the signature - requiring us to create a class with `new` keyword
- `tap-parser` added a parser.fullname field - 
https://github.com/tapjs/tap-parser/commit/287e137f0dc5854acc636d1f0f46f4ff423db4b4 - my approach in this PR is to explicitly delete the `parser.fullname` property, but I can change the *expectations* too if that's more preferred.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
         "through": "^2.3.8"
     },
     "devDependencies": {
-        "concat-stream": "^1.6.2",
+        "concat-stream": "^2.0.0",
         "eclint": "^2.8.1",
         "ecstatic": "^4.1.4",
         "eslint": "^6.8.0",
@@ -51,7 +51,7 @@
         "tap-parser": "^3.0.5"
     },
     "scripts": {
-        "prelint": "eclint check",
+        "prelint": "eclint check **/*.js",
         "lint": "eslint . bin/*",
         "pretest": "npm run lint",
         "test": "npm run tests-only",


osdevisnot:
Looking at the test failures, we might need to update the supported node version based on dependencies. 

Notably, `tap` started depending on `async`/`await`, `const` and other modern javascript features, which would require us to bump supported node version from ~1.0 to ~6.0 at least.

ljharb:
We will likely never drop any version of node that's currently supported. There's a reason we haven't upgraded all of the dev deps :-)

.gitignore
```diff
@@ -1,6 +1,9 @@
 # gitignore
 
 /node_modules
+/.idea
```
ljharb, line 4: please revert this file; these sorts of things should be in your personal global gitignore, not in every project you touch.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-version: ~> 1.0
+version: ~> 6.0
```
ljharb, line 1: what's the difference in travis yml versions?

package.json
```diff
@@ -41,17 +41,17 @@
...
         "falafel": "^2.2.4",
         "js-yaml": "^3.13.1",
-        "tap": "^8.0.1",
-        "tap-parser": "^3.0.5"
+        "tap": "^14.10.7",
+        "tap-parser": "^10.0.1"
     },
     "scripts": {
-        "prelint": "eclint check",
+        "prelint": "eclint check index.js lib/* bin/* test/* example/*",
```
ljharb, line 54: are those files not checked when no pattern is provided? also, would this work?
```suggestion
        "prelint": "eclint check '**/*.js'",
```

.gitignore
osdevisnot: fixed in latest commit.

package.json
osdevisnot: fixed in latest commit

osdevisnot:
> We will likely never drop any version of node that's currently supported. There's a reason we haven't upgraded all of the dev deps :-)

The motivation behind this PR is to be able to use the parallel testing feature introduced in tap version 13.

If I may ask, is there a specific reason to support node version if it reaches [EOL status on `node.js`](https://nodejs.org/en/about/releases/)

ljharb:
EOL status is irrelevant. People use it, and supporting it makes their platform upgrades easier.

osdevisnot:
ok, makes sense. Feel free to close this PR unless we want to change the support policies.

ljharb:
@osdevisnot i'm still interested in some of the changes here; if you can get concat-stream updated to v2, and improve the editorconfig check, that's still worth getting in.

osdevisnot:
closing this in favor of #506.

ljharb:
Looks like the concat-stream update isn't going to work in node < 4.

ljharb:
As such, I'm going to close this and #506; since it turns out, every non-updated dev dep was because updating it would break things :-)

PR #506, opened 2020-04-26
upgrade concat-stream

osdevisnot:
as discussed in #505


ljharb:
Looks like the concat-stream update isn't going to work in node < 4.

ljharb:
As such, I'm going to close this and #505; since it turns out, every non-updated dev dep was because updating it would break things :-)

PR #516, opened 2020-05-19
Update RegExp for matching stack frames to handle Promise/then scenario (fixes #515).

### LINKED ISSUE #515: Test.prototype._assert fails to identify source file/line for calls to assert methods inside Promise/then context
```bash
pi@claw:~/tape-issue $ node --version
v12.16.3
pi@claw:~/tape-issue $ npm install tape
+ tape@5.0.0
pi@claw:~/tape-issue $ cat tape-issue.js 
const tape = require("tape");

tape("repro", (test) => {
  test.plan(2);
  test.ok(false, "This call identifies the file/line");
  new Promise((resolve) => {
    resolve();
  }).then(() => {
    test.ok(false, "This call does NOT identify the file/line");
    test.end();
  });
});
pi@claw:~/tape-issue $ node tape-issue.js 
TAP version 13
# repro
not ok 1 This call identifies the file/line
  ---
    operator: ok
    expected: true
    actual:   false
    at: Test.<anonymous> (/home/pi/tape-issue/tape-issue.js:5:8)
    stack: |-
      Error: This call identifies the file/line
          at Test.assert [as _assert] (/home/pi/tape-issue/node_modules/tape/lib/test.js:260:54)
          at Test.bound [as _assert] (/home/pi/tape-issue/node_modules/tape/lib/test.js:84:32)
          at Test.assert (/home/pi/tape-issue/node_modules/tape/lib/test.js:379:10)
          at Test.bound [as ok] (/home/pi/tape-issue/node_modules/tape/lib/test.js:84:32)
          at Test.<anonymous> (/home/pi/tape-issue/tape-issue.js:5:8)
          at Test.bound [as _cb] (/home/pi/tape-issue/node_modules/tape/lib/test.js:84:32)
          at Test.run (/home/pi/tape-issue/node_modules/tape/lib/test.js:101:31)
          at Test.bound [as run] (/home/pi/tape-issue/node_modules/tape/lib/test.js:84:32)
          at Immediate.next [as _onImmediate] (/home/pi/tape-issue/node_modules/tape/lib/results.js:83:19)
          at processImmediate (internal/timers.js:456:21)
  ...
not ok 2 This call does NOT identify the file/line
  ---
    operator: ok
    expected: true
    actual:   false
    stack: |-
      Error: This call does NOT identify the file/line
          at Test.assert [as _assert] (/home/pi/tape-issue/node_modules/tape/lib/test.js:260:54)
          at Test.bound [as _assert] (/home/pi/tape-issue/node_modules/tape/lib/test.js:84:32)
          at Test.assert (/home/pi/tape-issue/node_modules/tape/lib/test.js:379:10)
          at Test.bound [as ok] (/home/pi/tape-issue/node_modules/tape/lib/test.js:84:32)
          at /home/pi/tape-issue/tape-issue.js:9:10
          at processTicksAndRejections (internal/process/task_queues.js:97:5)
  ...

1..2
# tests 2
# pass  0
# fail  2
```

The code inside the following loop seems to fail the `RegExp` match against the non-`tape` stack frames in the second case, so the `at:` field is missing from the output. (Specifically, the `res` object doesn't get its `functionName`, `file`, `line`, `column`, or `at` properties set.) An update of the `RegExp` to match lines like `at /home/pi/tape-issue/tape-issue.js:9:10` would seem to address this problem.

https://github.com/substack/tape/blob/0b5804d43068602e1615dfd395a3d85949bb03da/lib/test.js#L268-L328

DavidAnson:


DavidAnson:
Implemented this as requested in #515.

lib/test.js
```diff
@@ -299,9 +299,9 @@ Test.prototype._assert = function assert(ok, opts) {
                 Last part captures file path plus line no (and optional
                 column no).
 
-                    /((?:\/|[a-zA-Z]:\\)[^:\)]+:(\d+)(?::(\d+))?)/
+                    /((?:\/|[a-zA-Z]:\\)[^:\)]+:(\d+)(?::(\d+))?)\)?/
```
DavidAnson, line 302: This was outdated

lib/test.js
```diff
@@ -268,8 +268,8 @@ Test.prototype._assert = function assert(ok, opts) {
         for (var i = 0; i < err.length; i++) {
             /*
                 Stack trace lines may resemble one of the following. We need
-                to should correctly extract a function name (if any) and
```
DavidAnson, line 271: Removed extra "should"

test/anonymous-fn.js
```diff
@@ -20,7 +20,7 @@ tap.test('inside anonymous functions', function (tt) {
             'not ok 1 fail',
             '  ---',
             '    operator: fail',
-            '    at: Test.<anonymous> ($TEST/anonymous-fn/test-wrapper.js:$LINE:$COL)',
```
DavidAnson, line 23: This and the other changes like it resulted from the proposed change. As I read the stack in the test (below), the relevant line to report has always been the topmost one, though it wasn't used until now. Specifically in this case, the call to `t.fail` is happening from `anonymous-fn.js`, NOT `test-wrapper.js` as the test was previously asserting.

```
            '          at $TEST/anonymous-fn.js:$LINE:$COL',
            '          at Test.<anonymous> ($TEST/anonymous-fn/test-wrapper.js:$LINE:$COL)',
```

test/stackTrace.js
```diff
@@ -62,15 +62,18 @@ tap.test('preserves stack trace with newlines', function (tt) {
     });
 });
 
-tap.test('parses function name from original stack', function (tt) {
-    tt.plan(1);
+tap.test('parses function info from original stack', function (tt) {
+    tt.plan(4);
```
DavidAnson, line 66: I added assertions for `file`, `line`, and `column` to the existing tests.

test/stackTrace.js
```diff
@@ -98,6 +104,56 @@ tap.test('parses function name from original stack for anonymous function', func
...
+
+tap.test('parses function info from original stack for Promise scenario with anonymous function', function (tt) {
+    tt.plan(4);
+
+    var test = tape.createHarness();
+    test.createStream();
+
+    test._results._watch = function (t) {
+        t.on('result', function (res) {
+            tt.equal('<anonymous>', res.functionName);
```
DavidAnson, line 140: Fails without the change to `test.js`

lib/test.js
```diff
@@ -299,9 +299,9 @@ Test.prototype._assert = function assert(ok, opts) {
                 Last part captures file path plus line no (and optional
                 column no).
 
-                    /((?:\/|[a-zA-Z]:\\)[^:\)]+:(\d+)(?::(\d+))?)/
+                    /((?:\/|[a-zA-Z]:\\)[^:\)]+:(\d+)(?::(\d+))?)\)?/
             */
-            var re = /^(?:[^\s]*\s*\bat\s+)(?:(.*)\s+\()?((?:\/|[a-zA-Z]:\\)[^:\)]+:(\d+)(?::(\d+))?)\)$/;
+            var re = /^(?:[^\s]*\s*\bat\s+)(?:(.*)\s+\()?((?:\/|[a-zA-Z]:\\)[^:\)]+:(\d+)(?::(\d+))?)\)?$/;
```
DavidAnson, line 304: To be explicit, the new `?` groups with the `\)` to make that single trailing `)` optional

ljharb:
However, tests are failing in node 0.8 and 0.10 (and 0.6 and 0.9), which I think explains a bit about the way things are done in master :-/

test/stackTrace.js
```diff
@@ -98,6 +104,56 @@ tap.test('parses function name from original stack for anonymous function', func
     });
 });
 
+tap.test('parses function info from original stack for Promise scenario', function (tt) {
```
ljharb, line 107: this test needs to be skipped when `typeof Promise !== 'function'`

test/stackTrace.js
```diff
@@ -98,6 +104,56 @@ tap.test('parses function name from original stack for anonymous function', func
...
+        new Promise(function (resolve) {
+            resolve();
+        }).then(function onfulfilled() {
+            t.equal(true, false, 'true should be false');
+            t.end();
+        });
+    });
+});
+
+tap.test('parses function info from original stack for Promise scenario with anonymous function', function (tt) {
```
ljharb, line 132: this test needs to be skipped when `typeof Promise !== 'function'`

test/stackTrace.js
DavidAnson: Wrapped both `Promise` tests with the feature-detection code from `test/promises/fail.js`.

test/stackTrace.js
DavidAnson: Done

PR #524, opened 2020-08-17
Include name of test in log when test times out

ryanhamley:
It can be difficult to tell which test has timed out by looking at the logs since the log is hardcoded as `test timed out after...`. This PR includes the test name in the log to make it more explicit what timed out.


Raynos:
This looks great, thanks!

PR #530, opened 2020-09-22
fix: fixed package path not exported: ./bin/tape

Creplav:
I was having an error every time I would use a package that included tape regarding an export that would prevent me from continuing my pipelines.

The error was `Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './bin/tape' is not defined by "exports" in .../node_modules/tape/package.json`

This is a simple fix for that issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
         "./lib/results": "./lib/results.js",
         "./lib/test": "./lib/test.js",
         "./package": "./package.json",
-        "./package.json": "./package.json"
+        "./package.json": "./package.json",
+		"./bin/tape": "./bin/tape"
     },
     "bin": "./bin/tape",
     "directories": {


ljharb:
Can you elaborate on this? "exports" is only for things that are `require`d or `import`ed - the tape binary should only be executed, never required.

redbar0n:
The tape binary is `require`d downstream: https://github.com/ericelliott/riteway/blob/e647631c3abb4b05e1edb2bd69dde4d282072eef/bin/riteway#L3

Creplav:
> The tape binary is `require`d downstream by libraries building on tape: https://github.com/ericelliott/riteway/blob/e647631c3abb4b05e1edb2bd69dde4d282072eef/bin/riteway#L3

This is what I was referring to in the original post. When the test suite is run, it fails with the error above. Any time the package.json file is updated, this error comes back.

ericelliott:
@ljharb We use tape like a library and build on top of it in order to override the assertion API. I'm not sure how that could work if we just execute the bin.

```js
const withTape = tapeFn => (unit = '', TestFunction = noop) => tapeFn(unit, withRiteway(TestFunction));

// The testing library: a thin wrapper around tape
const describe = Object.assign(withTape(tape), {
  only: withTape(tape.only),
  skip: tape.skip
});
```

ljharb:
@ericelliott sure, but a binary executable is *not* a library and it's not reasonable to expect to use it like one.

You can execute the bin with `--require`, and patch tape itself before the tests run.

ericelliott:
workflow

ericelliott:
https://github.com/ericelliott/riteway/pull/293/commits/fe0b0a8efc65cc39d00154710b5a83711383e99c

ljharb:
i mean sure, that's one way to do it, but i fail to see why you wouldn't just `execSync` out to `tape` with a `--require` for the patches - that way you don't have to manually maintain copypasta.

Either way, since this has multiple solutions, I'll close this PR. Executables are not part of (anyone's) requireable/importable API.

ericelliott:
> i mean sure, that's one way to do it, but i fail to see why you wouldn't just execSync out to tape with a --require for the patches - that way you don't have to manually maintain copypasta.

Mainly because I don't have time to figure it out. 😆

PR #532, opened 2020-10-16
fix(lib/test): stack traces when res.name is Error, not string

petermetz:
When executing tests, the res.name property
(as constructed in the ./lib/test.js#_assert method)
is actually not a string but the error that was thrown.
This (assuming edge case) combined with
extra.error and opts.error both being falsy ends up
producing this situation where feeding res.name into
the Error constructor makes the original error's stack
trace disappear and only it's message remains with the
newly constructed Error instance holding a completely
generic stack trace that's internal to tape's source code
(and therefore not much use when a test is failing due to
some code is throwing exceptions instead of failing an
assertion).

Signed-off-by: Peter Somogyvari <peter.metz@unarin.com>

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -261,7 +261,8 @@ Test.prototype._assert = function assert(ok, opts) {
     this._ok = !!(this._ok && ok);
 
     if (!ok && !res.todo) {
-        res.error = defined(extra.error, opts.error, new Error(res.name));
+        var anError = res.name instanceof Error ? res.name : new Error(res.name);
+        res.error = defined(extra.error, opts.error, anError);
     }
 
     if (!ok) {


petermetz:
@ljharb Not sure why it's failing on older Node versions TBH. Do you have any pro tips where I could start with fixing the CI on NodeJS 7, 6, etc.?

Failing assertion here for example:
https://travis-ci.org/github/substack/tape/jobs/736511556#L1010

Thank you in advance!

ljharb:
@petermetz i've rebased this. it's passing ; but either way it needs a regression test.

petermetz:
> @petermetz i've rebased this. it's passing ; but either way it needs a regression test.

Thanks @ljharb I'll try to put a specific test case in there.

ljharb:
@petermetz any chance you're still planning to add the regression test?

petermetz:
@ljharb Apologies, still planning on it, yes. Might take some more time, but definitely have the intent. :-)

ljharb:
No worries, whenever you can get to it is fine.

ljharb:
@petermetz :-/ by deleting the fork, the PR is unrecoverable.

petermetz:
@ljharb Oh no, I'm sorry. I've completely forgot about this PR! I'll re-open from a new fork with a fresh PR that will include the regression test that you were requesting. Apologies once again, I was doing a big cleanup among my forks and somehow missed the pending PR on this one.

petermetz:
@ljharb I've taken a look at it again and it looks like in the meantime the same issue has been fixed elsewhere in the codebase. I haven't done a detailed investigation but my guess is that this is what I was intending originally as well: `./test/stackTrace.js` Thank you for being so responsive and sorry again for dropping the ball on this!

ljharb:
Alrighty, thanks for confirming.

PR #537, opened 2020-12-31
Preserve stacktraces for async/await

### LINKED ISSUE #536: Exceptions throw in `test('.', async (t) => { ... })` do not have meaningful stacktrace
When using `tape@5` instead of `tape@4` a programmer error in my test program is captured by `tape` and showing a `not ok ${err.message}`

```
# sync
not ok 1 TypeError: Cannot read property 'length' of undefined
  ---
    operator: fail
    stack: |-
      Error: TypeError: Cannot read property 'length' of undefined
          at Test.assert [as _assert] (/tape/lib/test.js:260:54)
          at Test.bound [as _assert] (tape/lib/test.js:84:32)
          at Test.fail (tape/lib/test.js:354:10)
          at Test.bound [as fail] (tape/lib/test.js:84:32)
          at onError (tape/lib/test.js:114:18)
  ...
```

The stack trace is the stack trace of tape itself handling `promise.catch()` ( https://github.com/substack/tape/blob/master/lib/test.js#L113 ) instead of the stack trace for my code.

I cannot tell which line of code in my test function has `.length of undefined` because tape is printing the wrong stack trace.

Ideally tape would rethrow the `err` to the global process uncaught exception handler so that I can see the real stack trace and so that the tests can fail hard and fast immediately.

Raynos:
We use `self.ifError(err)` instead of `self.fail()`
which logs the original stacktrace instead of creating
a new Error object.

Fixes #536

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -16,6 +16,7 @@ var isEnumerable = bind.call(Function.call, Object.prototype.propertyIsEnumerabl
 var toLowerCase = bind.call(Function.call, String.prototype.toLowerCase);
 var isProto = bind.call(Function.call, Object.prototype.isPrototypeOf);
 var $test = bind.call(Function.call, RegExp.prototype.test);
+var objectToString = bind.call(Function.call, Object.prototype.toString);
 
 module.exports = Test;
 
@@ -111,7 +112,11 @@ Test.prototype.run = function () {
                 self.end();
             }
         })['catch'](function onError(err) {
-            self.fail(err);
+            if (err instanceof Error || objectToString(err) === '[object Error]') {
+                self.ifError(err);
+            } else {
+                self.fail(err);
+            }
             self.end();
         });
         return;
--- a/test/async-await.js
+++ b/test/async-await.js
@@ -217,9 +217,10 @@ tap.test('async-error', function (t) {
             'ok 1 before throw',
             'not ok 2 Error: oopsie',
             '  ---',
-            '    operator: fail',
+            '    operator: error',
             '    stack: |-',
-            '      Error: Error: oopsie',
+            '      Error: oopsie',
+            '          at Test.myTest ($TEST/async-await/async-error.js:$LINE:$COL)',
             '          [... stack stripped ...]',
             '  ...',
             '',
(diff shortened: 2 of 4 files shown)


lib/test.js
```diff
@@ -111,7 +112,11 @@ Test.prototype.run = function () {
                 self.end();
             }
         })['catch'](function onError(err) {
-            self.fail(err);
+            if (objectToString(err) === '[object Error]') {
```
ljharb, line 115: I think this will only work for Error, not for the 6 builtin subclasses nor for user subclasses.

Raynos: do you have a preferred `is-error` library ?

I can also do `err instanceof Error || objectToString(err) === '[object Error]'`

`instanceof` should work for everything except cross realm which rarely happens ...

Raynos:
Addressed feedback.

lib/test.js
ljharb: Given that there's no robust way to identify Errors, cross-realm or otherwise, this seems like it'd be good enough.

test/async-await.js
```diff
@@ -245,3 +246,49 @@ tap.test('async-error', function (t) {
         t.end();
     });
 });
+
+tap.test('async-bug', function (t) {
+    runProgram('async-await', 'async-bug.js', function (r) {
+        var stdout = r.stdout.toString('utf8');
+        var lines = stdout.split('\n');
+        lines = lines.filter(function (line) {
+            return ! /^(\s+)at(\s+)<anonymous>$/.test(line);
```
ljharb, line 255: can you elaborate on why these lines need filtering out?

Raynos: I guess they don't that's just copy paste from a different test block, i glossed over it.

Raynos: Ok, removed the copy/paste stuff.

Raynos:
Cleaned up the test, this looks good to merge to me.

ljharb:
I’m going to land #538 first and rebase this, so i can be sure tests pass in all relevant node versions, and then I’ll land this one.

ljharb (approved):

ljharb:
I'll fix these last failing tests, and land this tonight.

Issue #540, opened 2021-01-07
t.throws() expected and msg parameters don't seem to behave as they are described.


cagross:
Hello.  I am a little confused by the documentation for [t.throws()](https://github.com/substack/tape#tthrowsfn-expected-msg), as it seems to be discrepant in two ways with the behavior I experience:

1.  The documentation says that the `expected` parameter, if present, must be a RegExp or Function.  But when I pass in a string, and execute a test, the test seems to complete without issue (see below for code).  Furthermore, the string passed as the `expected` parameter is printed to the console, seemingly as the test description. Nor does the `expected` string match the string representation of the exception (as described in the documentation).  [Here](https://i.imgur.com/Ju8HqXU.jpg) is a screenshot of a passing test--something similar occurs for a failing test.  Why is that?  What am I misunderstanding?

2.  The documentation mentions that the third parameter--`msg`--is an optional description of the assertion.  So when/where exactly is this string used?  I’ve executed two tests—one passing and one failing (see below for code).  In neither case is the `msg` parameter printed to the console (see [pass screenshot here](https://i.imgur.com/Ju8HqXU.jpg); see [fail screenshot here](https://i.imgur.com/P2wzdf4.jpeg)).
___
My passing test.  To change it to a failing test, comment out the `throw` line.

```
  t.throws(
    () => {
      throw 'Error message.'// Comment this line to force a failing test.
    },
    'Message 1',
    'Message 2'
  )
```

Thanks in advance.

ljharb:
Looking at the code, if `expected` is a string, *and* the error's `.message` property is equal to it, it fails with `The "error/message" argument is ambiguous.`. However, it shouldn't be possible for `expected` to be a string, because when it is a string, `expected` becomes `undefined` and `msg` becomes whatever the second argument is. In other words, in your example, `Message 1` becomes the message, `Message 2` becomes `extra` which as a string is unused - which means your assertion will pass as long as it throws anything, and fail as long as it doesn't.

It seems like it might be helpful here to fail if `extra` is a string, since that can't possibly do anything valuable, but I'm concerned that'd be a breaking change.

cagross:
OK thanks for all that--it is helpful.  

> However, it shouldn't be possible for expected to be a string, because when it is a string, expected becomes undefined and msg becomes whatever the second argument is. In other words, in your example, Message 1 becomes the message, 

[...]

PR #541, opened 2021-01-18
[readme] improve `t.throws`/`t.doesNotThrow` documentation

cagross:
Hi Jordan,

Sorry for the delay on this.  Per our discussion (in issue #540), in this commit I have edited the README to describe the following:

* The behavior of `t.throws()` when the `expected` parameter is a string.
* The fact that the `expected` parameter for `t.throws()` can be a validation object (just like Node's `assert.throws()`).

Have a look and let me know if anything is incomplete or inaccurate.

For now, due to time constraints, I omitted an example of validation object usage, and simply added a link to the Node documentation.  I'd very much like to add such an example (as they've done on [the Node documentation page](https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message)), but I'm out of time today, and might get busy with work tomorrow for the next week or so.  I can try to keep you posted here on that schedule. But it may be the case where you want to merge these edits now, and I open a new pull request next week which adds the validation object example.

Two more questions that we might want to address now (or earmark for future discussion):

1.  Are there files other than the README that should be updated with this information?  If so, I can do that if you want (either in this pull request or a separate pull request in the future).

2.  In the README, should other methods be updated with similar information?  i.e. are there other methods with an `expected` parameter that cannot be a string?  If so, I can do that if you want (either in this pull request or a separate pull request in the future).

--Carl


ljharb (approved):
Thanks, this looks great! Mind making the same changes to `doesNotThrow`, just below?

cagross:
> Mind making the same changes to doesNotThrow, just below?

OK sure.  Give me 24 hours and I'll push a new commit to this PR.

> (There aren't any other files to update; and other than throws/doesNotThrow, let's save additional updates for another PR)

OK noted on all.

cagross:
OK I just pushed a second commit with the edits to `t.doesNotThrow()`.  I omitted the validation object as an allowed value for `expected`, since Node's doesNotThrow does not allow it as a parameter.  I hope that was correct.  If not let me know and I can add it back in.

I also added some of the existing verbiage from `t.doesNotThrow()` to `t.throws()` because it looked helpful (speciifclaly the regex usage example).  Let me know if I did so incorrectly.

cagross:
Great!  Give me a few more days and I'll add that validation object example we discussed.  

If new issues are submitted related to thisarea, feel free to tag me (or this PR, or my commits, or whatever), and I can take a look.

PR #542, opened 2021-01-26
[readme] improve `t.throws` documentation

cagross:
* Added usage example for when expected parameter is a validation object.

Addresses #540.


ljharb (approved):
looks great!

cagross:
Great, thanks for allowing me to help :-)  Again, if there are issues with that documentation, or it needs modification/augmentation, feel free to ping me and I'll be happy to take a look.

yinsang:
Wrong case in doc.
I run the second and get passed.
```
const tape = require('tape')
tape('test for error', (t) => {
    t.plan(1)
    const err = new TypeError("Wrong value");
    err.code = 404;
    err.check = "true";

    // Failing test.
    t.throws(
        () => {
            throw err;
        },
        {
            code: 404,
            check: true // This is not deeply strictly equal to err.check.
        },
        "Test message."
    );
})
```
result is 
```
1..1
# tests 1
# pass  1

# ok
```
I tried to debug in source code,  [this line](https://github.com/substack/tape/blob/master/lib/test.js#L680) is wrong.
```
this._assert(typeof fn === 'function' && passed, {
        message: defined(msg, 'should throw'),
        operator: 'throws',
        actual: caught && caught.error,
        expected: expected,
         // while caught is an error object,  !passed return false, no error will be throw.
        error: !passed && caught && caught.error,
        extra: extra
    });
```

yinsang:
had fixed in the latest version. Sorry

銀桑 <notifications@github.com> 于2021年2月26日周五 下午10:53写道：

cagross:
YES :smiley:  I was _just_ posting here that I had a similar issue when I was testing that particular case--every test kept passing, even those that should fail.  Then I realized my version of tape was 4.14.4, and when I updated to v5+, the issue went away.

So you're not the only one it fooled.  Actually, maybe I should add a note to the README warning others of that.

ljharb:
I don’t think we need a note in the readme for something that applies to every software project ever written - “check the docs that match the version you’re using”.

PR #546, opened 2021-02-20
[New] add `.teardown()` on `t` instances

### LINKED ISSUE #531: Add beforeAll(), afterAll(), beforeEach(), afterEach()
Hello! I author and maintain many tape tests, and have found over the years that I frequently find a lot of setup, teardown, and repeat boilerplate code in these test suites. This could be greatly mitigated with the addition of optional `beforeAll()`, `afterAll()`, `beforeEach()`, `afterEach()` methods. (Happy to point to some examples if use cases are desired, although these kinds of features have become increasingly popular in other test libs.)

I'd propose that each of those methods run like any other test, but (as their names imply) be automatically inserted into the test flow before/after all tests in a given file (`beforeAll()`, `afterAll()`), and before/after each test in a given file (`beforeEach()`, `afterEach()`).

Unlike the module preloader and `t.onFinish()`, which are nice pre/post-run lifecycle hooks, I'd propose `beforeAll()`, `afterAll()`, `beforeEach()`, and `afterEach()` be incorporated into userland testing, and perform their own assertions, etc.

Example (pseudo):
```
test.beforeAll('Set up & test env')
test.afterAll('Tear down & test env')
test.beforeEach('Start service for each test')
test.afterEach('End service for each test')
test('Run a test')
test('Run another test')
test.onFinish('Run onFinish hook')
// Runs:
// - Set up & test env
// - Start service for each test
// - Run a test
// - End service for each test
// - Start service for each test
// - Run another test
// - End service for each test
// - Tear down & test env
// - Run onFinish hook
```

As alluded to in the example above, another reason I'd toss out in favor of adding this functionality is difficulty in maintaining large test suites where services or environments need to be set up and torn down at the beginning/end of the run. In that scenario, should one need to debug and isolate a single test, `t.only()` is not usable, because you still rely on the env setup/teardown tests; in such cases, I often find myself commenting out large blocks of code above and below the test I'm isolating, which is time consuming, error prone, and feels like a generally kind of janky workflow. (I assume I'm not completely alone in that, but who knows!)

Anyhow, happy to take a look at a PR with some guidance about if / how y'all would like to see it executed. Given that this is kind of out of band with how tape works today, would definitely want to get some feedback on how this would be best implemented.

mcollina:
Fixes #531

### DIFF
--- a/.github/workflows/node-zero.yml
+++ b/.github/workflows/node-zero.yml
@@ -12,6 +12,7 @@ jobs:
       - uses: ljharb/actions/node/matrix@main
         id: set-matrix
         with:
+          versionsAsRoot: true
           preset: '0.x'
 
   stable:
@@ -20,7 +21,8 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
-      matrix: ${{ fromJson(needs.matrix.outputs.stable) }}
+      matrix:
+        node-version: ${{ fromJson(needs.matrix.outputs.stable) }}
 
     steps:
       - uses: actions/checkout@v2
@@ -39,7 +41,10 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
-      matrix: ${{ fromJson(needs.matrix.outputs.unstable) }}
+      matrix:
+        node-version: ${{ fromJson(needs.matrix.outputs.unstable) }}
+        exclude:
+          - node-version: '0.6'
 
     steps:
       - uses: actions/checkout@v2
--- a/lib/test.js
+++ b/lib/test.js
@@ -69,6 +69,7 @@ function Test(name_, opts_, cb_) {
     this._plan = undefined;
     this._cb = args.cb;
     this._progeny = [];
+    this._teardown = [];
     this._ok = true;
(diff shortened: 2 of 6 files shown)


mcollina:
@ljharb I don't think we need an ending state, I did a few checks and I don't think we could trigger that situation.

Let me know if anything is missing.

ljharb:
Updated the readme, and added in a few more test cases (and fixes for those) inspired by https://github.com/avajs/ava/pull/2459 - non-function teardowns will now fail the test, and an error inside a teardown won't stop successive teardowns from running.

ljharb (approved):

ljharb:
The only failing test is node v0.6, which should have been failing since #318, but of course, tests weren't running on 0.6 at the time; and it also wasn't tested in the last travis build: https://travis-ci.org/github/substack/tape/builds/739518509, so the real confusion is why it wasn't failing in #538.

I'm just going to exclude v0.6 for now, since the last version of tap that supported v0.6 was a very long time ago.

mcollina:
wow. node v0.6!

ljharb:
@mcollina yup, i basically support everything forever :-)

ljharb:
Released in v5.2.0

PR #547, opened 2021-02-22
Use import() on esm files in supported node versions

lohfu:
This is a proof-of-concept of using `import()` to load ESM files in supported node versions (>= 14, >= 13.2.0, >= 12.17.0). The syntax breaking `import()` call is located in a separate file from `bin/tape.js` that is conditionally loaded, preventing older Node versions from crashing. 

This is currently an unfinished draft. I'm mainly curious to see if you are interested in something like this.

Currently this branch does not use promises to sequentially load the imported files. I have a separate branch, <https://github.com/lohfu/tape/blob/esm-support-promises/bin/load-module-esm-support.js#L27-L29>, where I have implemented sequential loading but this causes a test repo with a couple of simple test files to fail.

## Questions / Caveats

1. What is needed for the sequential loading to work? It seems like tape finishes before all tests are run properly. In a very simple repo with 2 test files all tests in the second file error with `not ok 2 test exited without ending: file 2 test 1`. The tests simply contain a single t.pass and a t.end. All tests in the tape repo pass.
2. Currently only `${cwd}/package.json` is loaded as I was a little reluctant to add a new dependency to find closest package.json... Is this ok? Do you have any preferences?
3. Is there a better way to determine ESM support than checking the node version?
4. nyc instrumentation fails: `failed to instrument /home/lohfu/dev/tape/bin/load-module-esm-support.js with error: SyntaxError: 'import' and 'export' may only appear at the top level (14:8)`... Can I update tap to get a newer nyc?
5. Currently only .mjs and .js files are loaded using `import()`. This prevents the use of loaders on other file extensions. In the future an --extension argument will probably be needed in the cli to allow for loaders on custom extensions.
6. Is the bin file ever bundled into browsers? I presume bundlers will add the file with the import to the bundle, which will cause older browsers to SyntaxError.


bin/load-module-esm-support.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var fs = require('fs');
+var extnamePath = require('path').extname;
+var resolvePath = require('path').resolve;
+
+var cwd = process.cwd();
+var pkgPath = resolvePath(cwd, 'package.json');
+
+var isModulePackage = fs.existsSync(pkgPath) && require(pkgPath).type === 'module';
```
ljharb, line 10: package.json files with `type: module` can be nested, and in particular, a package.json of `type: commonjs` can undo this, so the check has to always be relative to the `file`, and has to go all the way up the filesystem until it finds a `type` property.

bin/tape
```diff
@@ -41,6 +41,18 @@ if (typeof opts.ignore === 'string') {
     var matcher = ignore.createMatcher(ignoreStr);
 }
 
+function supportsEsm() {
+    var version = process.version.match(/^v(\d+)\.(\d+)\.\d+$/);
+    var major = parseInt(version[1], 10);
+    var minor = parseInt(version[2], 10);
+
+    return major >= 14 || major === 13 && minor >= 2 || major === 12 && minor >= 17;
```
ljharb, line 49: if we're trying to truly detect ESM, we'd have to determine if `import()` is a parsing error - but `import()` parses in node v13.2-v13.6 even though it can't work - `import()` always rejects in these versions.

If we're trying to check where it can work, then let's use https://www.npmjs.com/package/has-package-exports which already does this check properly.

ljharb (commented):
I agree that it's important these be serial - so when `import()` is supported, i think the entire thing should be a `.reduce` using Promises.

ljharb:
We also can’t upgrade tap due to the node versions it drops.

bin/tape
lohfu: i guess if testing for parsing errors can give false positives the safest would be to just use the second option... however it seems package exports (conditional) were unflagged in 12.17 and 13.7, while esm was unflagged in 12.17 and 13.2. just tested and `import()` works for me in 13.2.0, so this would lead to false negatives for node 13.2 - 13.6.

bin/load-module-esm-support.js
lohfu: will a nested package.json without a `type` "undo" a parent package.json with `type: module`?

lohfu:
the new tests i have added fail due to the problem i explained in the PR description, namely all tests in the files loaded after the first fail with `not ok 2 test exited without ending: module-b`

> We also can’t upgrade tap due to the node versions it drops.

hmmm, ok. any other way of circumventing the instrumentation error?

bin/load-module-esm-support.js
ljharb: No, only if it has an explicit type commonjs

bin/tape
ljharb: Works how? In node v13.2, when i `import()` an mjs file, i get "Error: Cannot find module".

ljharb:
I'm not sure what the instrumentation error is; is it about `import()`?

bin/tape
lohfu: This is most strange. I used nvm to install versions 13.2.0 - 13.6.0 and they all seem to work. My rudimentary test involved first creating the following files:

```js
// module-a.mjs
console.log(process.version)
import('./module-b.mjs').then(console.log)

// module-b.mjs
export default 1337
```

Running `$ node ./module-a.mjs` with node 13.2.0 prints
```
(node:481507) ExperimentalWarning: The ESM module loader is experimental.
v13.2.0
[Module] { default: 1337 }
```

lohfu:
regarding the failing tests, if you point me in the general direction of possible causes i am keen to start debugging and looking for solutions.

> I'm not sure what the instrumentation error is; is it about `import()`?

yes, the exact error is:

```
failed to instrument /home/lohfu/dev/tape/bin/load-module-esm-support.js with error: SyntaxError: 'import' and 'export' may only appear at the top level (16:8)
```

ljharb:
add `bin/load-module-esm-support.js` to `exclude` in `.nycrc`?

ljharb:
also we'll need to make sure the tests that do imports are skipped in the appropriate node versions.

bin/tape
ljharb: hmm, you're right. maybe it just doesn't work in the repl.

in that case, we do want to include v13.2. For now, let's keep your check, but i'll make a proper package to detect it and we can use that once it's available.

bin/load-module-esm-support.js
lohfu: i created a basic proof-of-conecpt implementation of traversing up the file tree to find a package.json with a type field (still need to verify this works in windows). i refrained from using for example [find-up](https://github.com/sindresorhus/find-up) since it uses too modern syntax. where should i place tests for this? next to the other, regular tape tests?

bin/load-module-esm-support.js
```diff
@@ -0,0 +1,42 @@
...
+var fs = require('fs');
+var path = require('path');
+var dirnamePath = path.dirname;
+var extnamePath = path.extname;
+var resolvePath = path.resolve;
+
+var cwd = process.cwd();
+var isModulePackage = getPackageType() === 'module';
+
+function getPackageType() {
```
ljharb, line 12: can we use https://npmjs.com/get-package-type for this?

bin/load-module-esm-support.js
```diff
@@ -0,0 +1,42 @@
...
+    var dir = __dirname;
+
+    while (dir !== '/' && dir !== process.env.HOME) {
```
ljharb, line 15: what does this do on windows?

bin/load-module-esm-support.js
lohfu: that package treats an empty type as commonjs, which i understood from you does not undo a parent package.json's `type: module`...

bin/load-module-esm-support.js
lohfu: not sure, once i know for sure this implementation will be used i'll set node up on windows and test it out

bin/load-module-esm-support.js
coreyfarrell: The goal of `get-package-type` is to determine if node.js will interpret `/path/to/file.js` as a module or commonjs.  Any discrepancy is a bug which I would want to fix.

bin/load-module-esm-support.js
```diff
@@ -0,0 +1,42 @@
...
+var path = require('path');
+var dirnamePath = path.dirname;
+var extnamePath = path.extname;
+var resolvePath = path.resolve;
+
+var cwd = process.cwd();
+var isModulePackage = getPackageType() === 'module';
+
+function getPackageType() {
+    var dir = __dirname;
```
coreyfarrell, line 13: It looks like you are determining the `type` which applies to `node_modules/tape/bin/load-module-esm-support.js`.  The reason `get-package-type` takes a filename argument is because each directory can be different even within a package.

bin/load-module-esm-support.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    return 'commonjs';
+}
+
+module.exports = function (file) {
+    var ext = extnamePath(file);
+
+    if (ext === '.mjs' || isModulePackage && ext === '.js') {
+        return import(resolvePath(cwd, file));
```
coreyfarrell, line 38:
```suggestion
        return import(pathToFileURL(resolvePath(cwd, file)));
```

IIRC Windows has issues if you pass a filesystem path to `import`, might be the `:` in absolute paths that causes problems.

bin/load-module-esm-support.js
coreyfarrell: The cross platform check I would use is `dir !== path.dirname(dir)`.

Note I would use `do { ... } while (dir !== path.dirname(dir))` - technically if package.json is in the root directory you would want to look at it.  Additionally I'm not aware of any limitation where node.js will not look at/above the home directory.

bin/tape
coreyfarrell: repl in earlier versions resolved imports relative to the parent directory.

bin/load-module-esm-support.js
lohfu: awesome, it would be much nicer to use your library instead of my subpar solution!

bin/load-module-esm-support.js
lohfu: this was meant to be cwd, not __dirname... checking for every file felt like unneeded extra work but it probably needs to be done. and i see your code caches the results so shouldnt be too much overhead.

bin/load-module-esm-support.js
lohfu: ugh, this is why we cant have nice things

bin/load-module-esm-support.js
lohfu: i just tested, it appears that a `package.json` without a "type" field does overwrite a parent "type: module". Ie if the closest `package.json` is missing the "type" field, the file is interpreted as commonjs.

lohfu: i just replaced my implementation with get-package-type... however i'm beginning to wonder if we should not just always use `import()` and let node determine esm vs cjs loading internally. @ljharb are there any downsides to always using `import()`?

ljharb: I suppose the only downside would be unnecessary asynchrony.

coreyfarrell: Would this cause spurious warnings about ES modules being experimental on certain versions of node.js?

ljharb: Yes, but those could be suppressed.

test/import.js
```diff
@@ -0,0 +1,71 @@
...
+            '',
+            '1..3',
+            '# tests 3',
+            '# pass  3',
+            '',
+            '# ok'
+        ].join('\n') + '\n\n');
+    };
+
+    var ps = tape('-r ./import/a import/test-a.js');
```
haroun, line 27: `import/test-a.js` does not exists, it should be `import/test-a.mjs` right or is it supposed to work anyway?

test/import.js
```diff
@@ -0,0 +1,71 @@
...
+            '',
+            '1..6',
+            '# tests 6',
+            '# pass  6',
+            '',
+            '# ok'
+        ].join('\n') + '\n\n');
+    };
+
+    var ps = tape('-r ./import/a -r ./import/b import/test-a.js import/test-b.js');
```
haroun, line 59: same here, should we use `.mjs` extension instead?

bin/tape
```diff
@@ -41,7 +41,19 @@ if (typeof opts.ignore === 'string') {
     var matcher = ignore.createMatcher(ignoreStr);
 }
 
-opts._.forEach(function (arg) {
+function supportsEsm() {
```
haroun, line 44: should we also check for `supportsEsm` in `opts.require` line 24?

I’m asking because in `test/import.js` we are passing `var ps = tape('-r ./import/a import/test-a.js'); ` where `./import/a` is a `.mjs` file.

It is leading to an hidden error (we’re only checking with `ps.stdout.pipe(concat(tc));`)
```txt
Error: Cannot find module 'test/import/a' from './tape' 
    at resolveSync (./tape/node_modules/resolve/lib/sync.js:108:15)
    at /home/harouna/repositories/tape/bin/tape:30:17
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/home/harouna/repositories/tape/bin/tape:24:14)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND'
}
```

I did the following as a proof of concept to make the tests go green

```patch
diff --git a/bin/tape b/bin/tape
index 4f5bfe9..20db7e1 100755
--- a/bin/tape
+++ b/bin/tape
@@ -22,12 +22,14 @@ if (typeof opts.require === 'string') {
 }
 
 opts.require.forEach(function (module) {
-    var options  = { basedir: cwd, extensions: Object.keys(require.extensions) };
+    var options  = { basedir: cwd, extensions: Object.keys(require.extensions).concat('.mjs') };
     if (module) {
         /* This check ensures we ignore `-r ""`, trailing `-r`, or
          * other silly things the user might (inadvertently) be doing.
          */
-        require(resolveModule(module, options));
+        supportsEsm()
+            ? require('./load-module-esm-support')(resolveModule(module, options))
+            : require(resolveModule(module, options));
     }
 });
```

haroun (changes requested):
@lohfu what do you think about the following changes? I guess it will not work in node versions without ESM support, but still I’ve tried it on node 14.15.3 and seems to make the tests go green. If you can help me understand what I’m missing, maybe I can do the changes. Thank you so much.

lib/results.js
```diff
@@ -81,8 +81,14 @@ Results.prototype.createStream = function (opts) {
         self._stream.pipe(output);
     }
 
-    if (!this._isRunning) {
-        this._isRunning = true;
+    if (!self._isRunning) {
+        self._isRunning = true;
+        Results.__loadingPromise ? Results.__loadingPromise.then(run) : run();
```
lohfu, line 86: @ljharb is exposing the loading promise this way an acceptable way of waiting for all tests to be loaded before running them?

lohfu:
@ljharb All tests pass now. If you are ok with the way i wait for all tests to load before running them I reckon the only things left now are:

1. Skipping import tests in appropriate versions
2. Deciding whether to check to use require or import for each file or always using import.

test/import.js
lohfu: yup, this was poor copy on my part. think i committed wrong as i initially had the tests passing locally. fixed now

lohfu:
Ok I did some very basic performance testing to see the implications of always using `import`. I created 10,000 empty `.js` files with a module type `package.json`. There was no observable performance benefit of removing the package type check and always using import. However, `require`:ing the same 10,000 files was around 4 times faster than importing them. Conclusion: no benefits when importing ESM files but serious disadvantage when importing CommonJS files.

lib/results.js
ljharb: It would be ideal not to expose it. Can this be a module-level variable instead?

lohfu: :thinking: not really sure what you mean by module-level variable here... the variable is set in `bin/tape.js`, but needs to accessed in `lib/results.js`

lohfu:
@ljharb what do you think about using this solution instead of checking which version of node is used to figure out if dynamic import is supported?

https://github.com/davidtheclark/cosmiconfig/pull/251/files#diff-e61e569af4ed5c2bd19c648e11225dd502296f0931b339c88cb82288a493e6ad

ljharb:
@lohfu that's basically the solution i'll be using in the package i forgot to create :-) let me get on that today.

ljharb:
@lohfu alrighty, give https://npmjs.com/has-dynamic-import a spin

lohfu:
> @lohfu alrighty, give https://npmjs.com/has-dynamic-import a spin

Sweet! Unfortunately I ran into some trouble with it immediately... I use has-dynamic-import to skip the ESM tests, and currently they arent skipped and fail in Node 10, 11, < 12.17 and < 13.2. All else seem to work fine. I created an [issue](https://github.com/inspect-js/has-dynamic-import/issues/2) and a [PR](https://github.com/inspect-js/has-dynamic-import/pull/3) explaining the issue and possible workaround.

lohfu:
i have fiddled around a little with getting this PR working with inspect-js/has-dynamic-import#3. all tests seem to be working on my end, but i have the following questions:

1. is the current way of handling async module loading ok? see:
  - https://github.com/substack/tape/pull/547/files#diff-cac4e8fa9811bc382c4e22b2d21dd2bf2b00675c458d17cbc221866db62f078eR55
  - https://github.com/substack/tape/pull/547/files#diff-75d9d2e71dcc077af46469ce4fce3f93e91f27b8f09b82dfa5574c304ee69aa3R86
2. should we have an option for custom extensions that enforces `import()` like `.mjs` does now? this would enable using loaders and other spooky magic (https://github.com/substack/tape/pull/547/files#diff-79dfc9fc6fa8fb5312506d39301093b519b95edbb72465a85102b1be7ec0f0f7R9)
3. should i squash/rebase away all "useless" commits or do you want the full history?

lohfu:
4. also, since full support check is now async and the tap `skip` option is sync, i used async sub-tests to exclude `import()` tests in unsupported environments (https://github.com/substack/tape/pull/547/files#diff-5ab063121df04838565bf235950a645d72c689af2409b48129368b6317bf23b2R8-R52)

lib/results.js
ljharb: Accessed how? Just reacting to the promise?

package.json
```diff
@@ -27,8 +27,10 @@
         "defined": "^1.0.0",
         "dotignore": "^0.1.2",
         "for-each": "^0.3.3",
+        "get-package-type": "^0.1.0",
         "glob": "^7.1.6",
         "has": "^1.0.3",
+        "has-dynamic-import": "github:lohfu/has-dynamic-import#handle-rejected-promise",
```
ljharb, line 33:
```suggestion
        "has-dynamic-import": "^2.0.0",
```

ljharb (commented):
Re 4, I think the `t.plan(2)` has to happen synchronously, otherwise tap has no idea that those tests are incoming.

Re 1, I'm not a fan of having `Results.__loadingPromise` exposed (see https://github.com/substack/tape/pull/547/files#r630197895)

Re 2, I'm not sure what you mean. "loaders" are already possible for CJS via `require.extensions`.

Re 3, please feel free to squash/rebase as needed.

lib/results.js
lohfu: yes

lohfu (commented):
> Re 4, I think the `t.plan(2)` has to happen synchronously, otherwise tap has no idea that those tests are incoming.

seemed to work fine, but i converted to using `t.end()` just in case.

> Re 1, I'm not a fan of having `Results.__loadingPromise` exposed (see https://github.com/substack/tape/pull/547/files#r630197895)

can't really think of any other solution... got any ideas?

> Re 2, I'm not sure what you mean. "loaders" are already possible for CJS via `require.extensions`.

I was referring to [ESM loaders](https://nodejs.org/api/esm.html#esm_loaders)... in the not too distant future i suspect some people want to `import()` files beyond `.mjs` or `.js`.

ljharb:
When node supports something more, we can add support for it. At the moment, there are no other kinds of modules.

lib/results.js
ljharb: So `bin/tape` creates the Promise, and `lib/results` needs to react to it?

Is there perhaps a way that either bin/tape could inject an executor into lib/results, or something?

lohfu: yes, precisely

well, `bin/tape` first needs to tell `lib/results` to wait and not immediately start the tests, and then it needs to say when they should be started.

perhaps we could implement a static function on Result to inject a promise instead of setting it manually?

```js
var loadingPromise = null;

Result.setLoadingPromise = function (promise) {
  loadingPromise = promise;
}

//

loadingPromise ? loadingPromise.then(run) : run();
```

that way would be more explicit.

bin/import-or-require.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var extnamePath = require('path').extname;
+var getPackageType = require('get-package-type');
+
+module.exports = function (file) {
+    var ext = extnamePath(file);
+
+    if (ext === '.mjs' || ext === '.js' && getPackageType.sync(file) === 'module') {
```
ljharb, line 9:
```suggestion
    if (ext === '.mjs' || (ext === '.js' && getPackageType.sync(file) === 'module')) {
```
for clarity

bin/tape
```diff
@@ -41,18 +33,43 @@ if (typeof opts.ignore === 'string') {
...
+             * other silly things the user might (inadvertently) be doing.
+             */
+            return module;
+        }).map(function (module) {
+            return resolveModule(module, { basedir: cwd, extensions: Object.keys(require.extensions) });
+        }).reduce(function (loadingPromise, module) {
+            return loadingPromise ? loadingPromise.then(function () {
+                return importOrRequire(module);
+            }) : importOrRequire(module);
+        }, undefined);
```
ljharb, line 53:
```suggestion
        }).reduce(function (loadingPromise, module) {
            return loadingPromise.then(function () {
                return importOrRequire(module);
            });
        }, Promise.resolve());
```

bin/tape
```diff
@@ -41,18 +33,43 @@ if (typeof opts.ignore === 'string') {
...
-        require(resolvePath(cwd, file));
-    });
+        return files
+            .filter(function (file) { return !matcher || !matcher.shouldIgnore(file); })
+            .map(function (file) { return resolvePath(cwd, file); })
+            .reduce(function (loadingPromise, file) {
+                return loadingPromise ? loadingPromise.then(function () {
+                    return importOrRequire(file);
+                }) : importOrRequire(file);
+            }, loadingPromise);
```
ljharb, line 71:
```suggestion
            .reduce(function (loadingPromise, file) {
                return loadingPromise.then(function () {
                    return importOrRequire(file);
                });
            }, loadingPromise || Promise.resolve());
```

bin/tape
```diff
@@ -41,18 +33,43 @@ if (typeof opts.ignore === 'string') {
...
-    if (!Array.isArray(files)) {
-        throw new TypeError('unknown error: glob.sync did not return an array or throw. Please report this.');
-    }
+    var loadingPromise = opts.require
+        .filter(function (module) {
+            /* This check ensures we ignore `-r ""`, trailing `-r`, or
+             * other silly things the user might (inadvertently) be doing.
+             */
+            return module;
+        }).map(function (module) {
```
ljharb, line 47:
```suggestion
         // This check ensures we ignore `-r ""`, trailing `-r`, or other silly things the user might (inadvertently) be doing.
        .filter(Boolean)
        .map(function (module) {
```

lib/results.js
ljharb: There might be a simpler way. What code calls Results.createStream in the first place? is there some way bin/tape could start the tests itself, and "pause" them, rather than needing static state and mutation?

bin/tape
lohfu: wont we loose support for Node 0.10 and earlier with this logic (they lack promise support)?

lib/results.js
lohfu: `createStream` is called in `createExitHarness`, https://github.com/substack/tape/blob/master/index.js#L62.

i tested adding a `wait` option to `getHarness`, and calling `tape.getHarness({ wait: true })` at the start of `bin/tape`. I also exposed a `tape.run()` method that creates the stream which i call when the loadingPromise has been resolved in `bin/tape`.

One negative side effect of this that would need to be fixed is that custom streams could not be created overriding the default stream as it will already have been created.

Another option would be to implement `tape.wait()` which set an a boolean and gets passed to `getHarness` internally.

lohfu: all those changes are in https://github.com/substack/tape/pull/547/commits/4f9787a71f259b179bef84ad4829dc51b5da380b

bin/tape
lohfu: fixed

ljharb (commented):
It occurred to me today that node only has `--require` and does not have any way to evaluate ESM files on process load - it might be better to wait until node has that support before adding it here, in case they conflict.

eg, it's possible node will add `--import=` for ESM, in which case that's what we'd want to have.

lohfu:
> It occurred to me today that node only has --require and does not have any way to evaluate ESM files on process load - it might be better to wait until node has that support before adding it here, in case they conflict.

removed!

lib/results.js
lohfu: since a bit of logic would have to be rethought to enable overriding the default stream after calling `getHarness`, i changed the logic to using a `wait` method and an internal wait boolean.

ljharb:
I'm not sure what you removed - what i mean is, `--require=./foo.mjs` is impossible in node, so i'm wondering if it should remain impossible in tape.

lohfu:
i removed the logic to use `import()` statements on ESM file in the `--require` option.

i am feeling pretty ready with this PR now... if you are ok with the `wait()`/`run()` logic for handling async importing of tests before running them i will update the README and squash everything to a single commit.

bin/tape
```diff
@@ -41,18 +46,31 @@ if (typeof opts.ignore === 'string') {
...
-    if (!Array.isArray(files)) {
-        throw new TypeError('unknown error: glob.sync did not return an array or throw. Please report this.');
-    }
+    var loadingPromise = opts._.reduce(function (loadingPromise, arg) {
+        // If glob does not match, `files` will be an empty array.
+        // Note: `glob.sync` may throw an error and crash the node process.
+        var files = glob.sync(arg);
+
+        if (!Array.isArray(files)) {
+            throw new TypeError('unknown error: glob.sync did not return an array or throw. Please report this.');
```
ljharb, line 60: this will now reject the Promise, instead of exiting the program

bin/tape
```diff
@@ -41,18 +46,31 @@ if (typeof opts.ignore === 'string') {
...
+                return loadingPromise ? loadingPromise.then(function () {
+                    return importOrRequire(file);
+                }) : importOrRequire(file);
+            }, loadingPromise);
+    }, undefined);
 
-    files.filter(function (file) { return !matcher || !matcher.shouldIgnore(file); }).forEach(function (file) {
-        require(resolvePath(cwd, file));
-    });
+    loadingPromise ? loadingPromise.then(tape.run) : tape.run();
```
ljharb, line 73:
```suggestion
    if (loadingPromise) {
        loadingPromise.then(function () { tape.run(); });
    } else {
        tape.run();
    }
```

index.js
```diff
@@ -92,6 +110,7 @@ function createExitHarness(conf) {
     });
 
     return harness;
+
```
ljharb, line 113:
```suggestion
```

lib/results.js
```diff
@@ -81,8 +81,9 @@ Results.prototype.createStream = function (opts) {
         self._stream.pipe(output);
     }
 
-    if (!this._isRunning) {
-        this._isRunning = true;
+    if (!self._isRunning) {
+        self._isRunning = true;
```
ljharb, line 85: why this change?

package.json
```diff
@@ -27,8 +27,10 @@
         "defined": "^1.0.0",
         "dotignore": "^0.1.2",
         "for-each": "^0.3.3",
+        "get-package-type": "^0.1.0",
```
ljharb, line 30: I’m a bit concerned that this package’s engines declaration will cause false errors for yarn users, and anyone using engines-strict or https://npmjs.com/ls-engines, if it ever adds an engines declaration in the future (cc @coreyfarrell)

bin/tape
```diff
@@ -8,6 +8,11 @@ var readFileSync = require('fs').readFileSync;
 var parseOpts = require('minimist');
 var glob = require('glob');
 var ignore = require('dotignore');
+var hasDynamicImport = require('has-dynamic-import');
+var tape = require('../');
+var Results = require('../lib/results');
```
ljharb, line 13:
```suggestion
```

ljharb (changes requested):
Definitely getting close!

lib/results.js
lohfu: not sure, left over after my revert. i checked out the file from a previous commit and committed now so should be fixed.

bin/tape
lohfu: i moved the glob function into a reducer outside the promise. i also updated the error message to read

```js
throw new TypeError('unknown error: glob.sync("' + arg + '") did not return an array or throw. Please report this.');
```

to make it easier to see which glob call failed.

bin/tape
```diff
@@ -41,18 +45,36 @@ if (typeof opts.ignore === 'string') {
...
 
-opts._.forEach(function (arg) {
+var files = opts._.reduce(function (result, arg) {
     // If glob does not match, `files` will be an empty array.
     // Note: `glob.sync` may throw an error and crash the node process.
     var files = glob.sync(arg);
 
-    if (!Array.isArray(files)) {
-        throw new TypeError('unknown error: glob.sync did not return an array or throw. Please report this.');
+    if (files[0] === 'test/import/mjs-e.mjs') {
```
ljharb, line 53: i assume this was just for testing

bin/tape
```diff
@@ -41,18 +45,36 @@ if (typeof opts.ignore === 'string') {
...
+            return importOrRequire(file);
+        }) : importOrRequire(file);
+    }, undefined);
+
+    if (loadingPromise) {
+        loadingPromise.then(tape.run);
+    } else {
+        tape.run();
+    }
 });
```
ljharb, line 78: thanks, this is great.

imagine, however, requiring a file that does `throw 42` - previously/currently, that would be an uncaught exception. what happens here?

index.js
```diff
@@ -48,25 +57,34 @@ exports = module.exports = (function () {
...
+        }
+        stream.on('end', function () { ended = true; });
+    };
 
-    var ended = false;
-    stream.on('end', function () { ended = true; });
+
+    if (!wait) {
+        harness.run();
+    }
```
ljharb, line 87: maybe when `wait` is falsy, `harness.run` shouldn't exist? iow, maybe this should be `if (wait) { harness.run = run; } else { run(); }`?

bin/tape
lohfu: argh, at least three times i looked at that and thought "better not commit" that. my bad

bin/tape
lohfu: in Node < 15, it returns a UnhandledPromiseRejectionWarning. In 15 and beyond it behaves just like the uncaught exception would.

If i add a simple catch to the promise, the errors from inside a promise loop will look slightly differently from completely sync uncaught exceptions:

Handled rejection:
```
Error: what
    at file:///home/notme/dev/tape/tmp.mjs:1:7
    at ModuleJob.run (node:internal/modules/esm/module_job:154:23)
    at async Loader.import (node:internal/modules/esm/loader:177:24)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:437:15)
```

Unhandled rejection in Node >=15

```
file:///home/lohfu/dev/tape/tmp.mjs:1
throw new Error('what')
      ^

Error: what
    at file:///home/lohfu/dev/tape/tmp.mjs:1:7
    at ModuleJob.run (node:internal/modules/esm/module_job:154:23)
    at async Loader.import (node:internal/modules/esm/loader:177:24)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:437:15)
```


Uncaught exception:
```
/home/notme/dev/tape/tmp.js:1
throw new Error('what')
      ^

Error: what
    at Object.<anonymous> (/home/notme/dev/tape/tmp.js:1:7)
    at Module._compile (node:internal/modules/cjs/loader:1092:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1121:10)
    at Module.load (node:internal/modules/cjs/loader:972:32)
    at Function.Module._load (node:internal/modules/cjs/loader:813:14)
    at Module.require (node:internal/modules/cjs/loader:996:19)
    at require (node:internal/modules/cjs/helpers:92:18)
    at module.exports (/home/notme/dev/tape/bin/import-or-require.js:12:9)
    at /home/notme/dev/tape/bin/tape:70:14
    at Array.reduce (<anonymous>)
```

index.js
lohfu: could be done, i would have to update `lazyload.run` a little bit to:

```js
lazyLoad.run = function () {
    var run = getHarness().run;
    
    if (run) {
        run();
    }
};
```

bin/tape
lohfu: I would preferably not loose the native unhandled rejection behaviour in Node >= 15, but conditionally adding a catch statement for older nodes feels like a smelly solution. what do you reckon?

index.js
ljharb: or `lazyload.run = getHarness().run || function () {};`?

bin/tape
ljharb: that behavior is horrific and violates the language spec, and i actively do not want to take advantage of it. An unhandled rejection is perfectly fine and should not inherently exit any program.

as such, we should be explicitly handling rejections in all cases, even if it's `.catch()` solely to silence node 15+.

index.js
lohfu: unfortunately that logic is called on `require('tape')`, so we will not be able to call `tape.wait()` before. it will also make it impossible to override the default stream.

bin/tape
lohfu: ok, fixed!

bin/tape
```diff
@@ -41,18 +45,42 @@ if (typeof opts.ignore === 'string') {
...
+            return importOrRequire(file);
+        }) : importOrRequire(file);
+    }, undefined);
+
+    if (loadingPromise) {
+        loadingPromise.then(
+          tape.run,
+          function (err) {
+            console.error(err);
+            process.exit(1);
```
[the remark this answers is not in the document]
lohfu: that unfortunately makes it more confusing, as the first output becomes

```
/home/notme/dev/tape/bin/tape:79
                setTimeout(function () { throw err; }, 0);
```

instead of saying where the error was originally thrown.

bin/tape
```diff
@@ -41,18 +45,42 @@ if (typeof opts.ignore === 'string') {
...
+          }
+        );
+    } else {
+        tape.run();
```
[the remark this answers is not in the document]
lohfu: yes, that works. however, it apparently fails if Node supports a promise, but no `import()` call has been made. since `hasDynamicImport` creates a promise, we need to move the catch outside. I had to rethink the logic a bit... i'm beginning to feel ashamed :disappointed: ... but changed and pushed.

i should probably add some tests for this error handling.

.eslintrc
```diff
@@ -29,6 +29,19 @@
...
+            "parserOptions": {
+                "ecmaVersion": 2020,
+                "sourceType": "module",
+            },
+        },
+        {
+            "files": ["bin/import-or-require.js"],
+            "parserOptions": {
+                "ecmaVersion": 2020,
+            },
```
ljharb, line 43:
```suggestion
            "extends": "@ljharb/eslint-config/node",
```

bin/tape
```diff
@@ -41,18 +45,44 @@ if (typeof opts.ignore === 'string') {
...
+    if (hasSupport) {
+        var importOrRequire = require('./import-or-require');
+
+        return files.reduce(function (promise, file) {
+            return promise.then(function () {
+                return importOrRequire(file);
+            });
+        }, Promise.resolve());
+    } else {
+        files.forEach(require);
```
ljharb, line 72:
```suggestion
        files.forEach(function (x) { require(x); });
```
let's be explicit, especially since forEach passes 3 arguments.

bin/tape
```diff
@@ -41,18 +45,44 @@ if (typeof opts.ignore === 'string') {
...
+    }
 });
 
+if (promise) {
```
ljharb, line 76: how could this ever be falsy?

lohfu: when there is no promise support, hasDynamicImport's thenable returns undefined. i will add a comment saying this.

.eslintrc
lohfu: :thinking: not sure if i'm missing something here, but the `node` directory in that repo only contains `latest.json` and no `index.js`... so eslintrc complains that the config cannot be found...

bin/tape
ljharb: ahhh. that seems like kind of a bug in hasDynamicImport - it'd be ideal not to rely on that.

lohfu: fair enough, i rewrote it to use `has-dynamic-import`'s syntax check before checking full support and going into promise mode

.eslintrc
ljharb: oops, right
```suggestion
            "extends": "@ljharb/eslint-config/node/latest",
```

ljharb (changes requested):
i've rebased the PR; please fetch before making further changes.

lohfu:
so i started using the changes in this branch on a project, and i've found a major drawback to the handling of rejections when loading test files.

https://github.com/substack/tape/blob/229bd7876ec71d5c8d7681a4f2e6c13cb61a0efe/bin/tape#L67-L73

catching errors in our loading promise, `console.error`:ing them and then exiting makes it very difficult to locate syntax errors in ESM files. running an ESM file with invalid syntax "natively" in Node prints:

```
$ node tmp.mjs
file:///home/lohfu/dev/tape/tmp.mjs:1
)
^

SyntaxError: Unexpected token ')'
    at Loader.moduleStrategy (node:internal/modules/esm/translators:147:18)
    at async link (node:internal/modules/esm/module_job:64:21)
```

the actual stack trace contains no information of where the syntax error is; it's the lines before that the discloses the error and its location. running the same file with this branch prints:

```
$ ./bin/tape tmp.mjs
SyntaxError: Unexpected token ')'
    at Loader.moduleStrategy (node:internal/modules/esm/translators:147:18)
    at async link (node:internal/modules/esm/module_job:64:21)
```

that information obviously makes it difficult to locate the problem without external tools. node appears to behave differently when the syntax errors are in commonjs files, ie when tape calls `require()` and not `import()`. in that case, `console.error(err)` prints the syntax error and it's location:

```
$ ./bin/tape tmp.cjs
/home/notme/dev/tape/tmp.cjs:1
)
^

SyntaxError: Unexpected token ')'
    at Object.compileFunction (node:vm:353:18)
    at wrapSafe (node:internal/modules/cjs/loader:1039:15)
    at Module._compile (node:internal/modules/cjs/loader:1073:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1138:10)
    at Module.load (node:internal/modules/cjs/loader:989:32)
    at Function.Module._load (node:internal/modules/cjs/loader:829:14)
    at Module.require (node:internal/modules/cjs/loader:1013:19)
    at require (node:internal/modules/cjs/helpers:93:18)
    at module.exports (/home/notme/dev/tape/bin/import-or-require.js:12:5)
    at /home/notme/dev/tape/bin/tape:85:24
```

ljharb:
Yeah, i think node does some magic with exceptions that are not caught and re-thrown. I’m not sure how to solve for that.

lohfu:
After giving it some thought I have come up with some changes. The basic idea is to use `process.nextTick` to leave the promise chain after checking for dynamic import support and only enter into a new promise chain if we need to load ESM files. This way we avoid unhandled rejections when only loading commonjs files, ie in all current setups using tape. If we do enter into a promise chain when loading files, errors are no longer caught. This makes node 15 onward handle errors in the loading promise very similar to how they are handled in sync loading. I know one should usually strive for handling all rejections, but i reckon this is a bit of an edge case. We want the program to quit when there are errors importing files, and we want to show exactly what Node natively prints. If rejections need to be handled when importing files, shouldn't the current `require` calls be wrapped in a try/catch?

After these changes, this PR will only affect people that consciously try to load ESM files. It will only be awkward for people that are trying to use ESM files in Node versions before 15. People converting their projects to full ESM most likely use very modern versions and as such, this should affect few.


On errors importing files, the previous changes printed (`console.error()`:ing rejections `import()`)::

```
SyntaxError: Unexpected token ')'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:84:18)
    at async link (internal/modules/esm/module_job.js:36:21)
```

For rejections `require()`:ing files, the output is the same as requiring the files outside the promise. 

With the changes from the latest commits, different Node versions will print something like:

Node >=15

```
file:///home/notme/dev/tape/tmp.mjs:1
)
^

SyntaxError: Unexpected token ')'
    at Loader.moduleStrategy (node:internal/modules/esm/translators:147:18)
    at async link (node:internal/modules/esm/module_job:48:21)

```

Node ^12.17, ^13.2 and 14 prints:

```
(node:91315) UnhandledPromiseRejectionWarning: SyntaxError: Unexpected token ')'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:84:18)
    at async link (internal/modules/esm/module_job.js:36:21)
(node:91315) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:91315) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

All Node versions that do not support `import`:

```
/home/notme/dev/tape/tmp.mjs:1
)
^

SyntaxError: Unexpected token )
    at Module._compile (internal/modules/cjs/loader.js:760:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Module.require (internal/modules/cjs/loader.js:723:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at /home/lohfu/dev/tape/bin/tape:86:34
    at Array.forEach (<anonymous>)
    at importFiles (/home/lohfu/dev/tape/bin/tape:86:11)
    at processTicksAndRejections (internal/process/task_queues.js:81:17)
```

ljharb:
> If rejections need to be handled when importing files, shouldn't the current require calls be wrapped in a try/catch?

It's not that they need to be handled; it's that module-level exceptions (at require/import time) should crash the process; unhandled rejections (including from an ESM module that does `export function then() { throw 42 }`) should not.

Ideally, a module-level exception from a CJS file should look the same in any version of node, whether or not you're including ESM files.

Your last comment implies we might be pretty close to the ideal scenario :-) I'll take a look!

lohfu:
Since playing around a little with the `--unhandled-rejections=mode` CLI option I am getting quite convinced this is a reasonable solution. Using the `NODE_OPTIONS` env variable to set unhandled rejections mode to either `strict` or `throw`, Node and tape handles unhandled rejections the same in all version of Node that supports ESM imports.


```
$ NODE_OPTIONS='--unhandled-rejections=throw' ./bin/tape tmp.mjs
file:///home/notme/dev/tape/tmp.mjs:1
)
^

SyntaxError: Unexpected token ')'
    at Loader.moduleStrategy (node:internal/modules/esm/translators:147:18)
    at async link (node:internal/modules/esm/module_job:48:21)
```

I added some tests to check how the process exits, but do not really have any tests checking the error output.

lohfu:
@ljharb have you had any time to review the latest changes?

ljharb (approved):

ljharb:
@lohfu thanks for your patience and for working so thoroughly on this, I really appreciate it!

I'll cut a release that includes this soon.

ljharb:
See v5.3.0

alvinometric:
Just here to say that this is amazing, thank you everyone 🥳

PR #554, opened 2021-05-04
[readme] Update the documention for t.doesNotThrow()

### LINKED ISSUE #553: t.doesNotThrow() expected parameters seems to be ignored
Hello,

I read the code of the method [`t.doesNotThrow()`](https://github.com/substack/tape/blob/v5.2.2/lib/test.js#L679) and I don't understand the purpose of the `expected` parameter.

The documentation says :
> `expected`, if present, limits what should not be thrown, and must be a `RegExp` or `Function`. The `RegExp` matches the string representation of the exception, as generated by `err.toString()`. For example, if you set `expected` to `/user/`, the test will fail  only if the string representation of the exception contains the word `user`. Any other exception will result in a passed test.  The `Function` is the exception thrown (e.g. `Error`). 

So, if my understanding is correct, the following tests should not fail :

```js
const test = require("tape");

test('should not fail', (t) => {
    t.doesNotThrow(
        () => { throw new Error("Foo bar") },
        TypeError,
        "the callback does not throw a TypeError"
    );
    t.end();
});

test('should not fail', (t) => {
    t.doesNotThrow(
        () => { throw new Error("Foo bar") },
        /baz/,
        "the callback does not throw 'baz'"
    );
    t.end();
});
```
But the current behavior of `t.doesNotThrow()` is to always fail if the callback throws an exception.

Should the method `t.doesNotThrow()` be updated to match the documentation, or is my understanding of the documentation incorrect ?

m-r-r:
Hello,

I updated the documentation for `t.doesNotThrow()` to clarify how it works.

Fixes #553.


ljharb:
Hmm. Your text is correct, but node's docs say that it does check the type - in terms of what kind of error is thrown. Since tape fails the test instead of throwing an error, it kind of seems like we should be failing the test when it doesn't throw the right kind of error, and instead, throwing the error when it throws the wrong kind of error?

ljharb:
@m-r-r unfortunately by deleting the fork, this PR is unrecoverable.

PR #556, opened 2021-07-02
Another way to create custom reporters

ydarma:
This is the way I do things and I wanted to share it if you think that is worth...

Thank you for your work

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -447,6 +447,20 @@ $ node object.js test/x.js test/y.js
 {"type":"end","test":2}
 ```
 
+A convenient alternative to achieve the same:
+```js
+// report.js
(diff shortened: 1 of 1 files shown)


readme.markdown
```diff
@@ -447,6 +447,21 @@ $ node object.js test/x.js test/y.js
 {"type":"end","test":2}
 ```
 
+An convenient alternative to achieve the same:
+```js
+// report.js
+var test = require('tape');
+var path = require('path');
```
Raynos, line 454: path variable is unused.

ydarma: You're absolutely right. I fixed this.
Thanks for your comment.

ljharb (approved):

PR #558, opened 2021-07-23
correct docs for t.teardown

laramackey:
Correction to README for the `teardown` method as it's on the `t` instance rather than `test`

See https://github.com/substack/tape/pull/546 for reference

Also open PR for updating type definition https://github.com/DefinitelyTyped/DefinitelyTyped/pull/54694


ljharb (approved):
Thanks!

PR #559, opened 2021-08-17
Correct esm problem on windows

David-Desmaisons:
When using tape on windows in a esm module (using `"type": "module"` in the project package.json),  this exception is raised:

```
(node:7484) UnhandledPromiseRejectionWarning: Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are
supported by the default ESM loader
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:698:11)
    at Loader.resolve (internal/modules/esm/loader.js:97:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:243:28)
    at Loader.import (internal/modules/esm/loader.js:178:28)
    at importModuleDynamically (internal/modules/cjs/loader.js:1080:27)
    at exports.importModuleDynamicallyCallback (internal/process/esm_loader.js:37:14)
    at importOrRequire (C:\...\tests\node_modules\tape\bin\import-or-require.js:11:9)       
    at C:\...\tests\node_modules\tape\bin\tape:84:14
    at Array.reduce (<anonymous>)
    at importFiles (C:\...\tests\node_modules\tape\bin\tape:81:30)
```
In turns out that making explicit the file protocol in the importOrRequire resolves this problem.

### DIFF
--- a/bin/import-or-require.js
+++ b/bin/import-or-require.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { extname: extnamePath } = require('path');
+const { pathToFileURL } = require('url');
 const getPackageType = require('get-package-type');
 
 // eslint-disable-next-line consistent-return
@@ -8,7 +9,7 @@ module.exports = function importOrRequire(file) {
     const ext = extnamePath(file);
 
     if (ext === '.mjs' || (ext === '.js' && getPackageType.sync(file) === 'module')) {
-        return import(file);
+        return import(pathToFileURL(file).href);
     }
     require(file);
 };


bin/import-or-require.js
```diff
@@ -8,7 +8,7 @@ module.exports = function importOrRequire(file) {
     const ext = extnamePath(file);
 
     if (ext === '.mjs' || (ext === '.js' && getPackageType.sync(file) === 'module')) {
-        return import(file);
+        return import("file://" + file);
```
ljharb, line 11: this will break if someone uses `--require=packageName`. You need to provide the `flle://` on relative ESM paths yourself, I think.

ljharb (changes requested):

David-Desmaisons:
I just made a test with the --require option and it seems that it is not calling the `importOrRequire` function:
image

Is it correct or I am missing something?

ljharb:
Sorry, you’re right; i meant `--import=packageName`

ljharb:
Also, on what node version are you seeing this issue? And what exact tape command are you running?

David-Desmaisons:
Here is a minimal repo for reproduction using `tape **/*.spec.js` command:
https://github.com/David-Desmaisons/tape-tests

I am testing with node 12.

ljharb:
You can’t pass unquoted globs to tools tho - your shell will expand them. Try putting quotes around the glob.’

If that doesn’t fix it, I’ll take a look at your repro repo shortly.

David-Desmaisons:
I changed to:
```
  "scripts": {
    "test": "tape \"**/*.spec.js\""
  },
```
With the same results: it works on ubuntu but breaks on windows.

ljharb:
interesting, ok. it's also notable you're not using `--import` or `--require`, which means this line of code shouldn't be executed. I don't have Windows to be able to test, unfortunately.

ljharb:
Indeed, your repo passes fine on my Mac.

David-Desmaisons:
The path are constructed line 59 of tape.js file

image

It is then used using the import sintax but as on windows it is a windows path it is not compatible with an url:

image

Adding "file:\\" before the path resolve the windows case without breaking linux build.

bin/import-or-require.js
David-Desmaisons: I am still not sure about the `--import=packageName` (only find the require and ignore option). Could you please give an example that would break with this fix?

bin/import-or-require.js
```diff
@@ -8,7 +8,7 @@ module.exports = function importOrRequire(file) {
     const ext = extnamePath(file);
 
     if (ext === '.mjs' || (ext === '.js' && getPackageType.sync(file) === 'module')) {
-        return import(file);
+        return import('file://' + file);
```
vweevers, line 11: I'm seeing the same issue on Windows (with node 16.9.1).

I suggest `import(pathToFileURL(file).href)`, using `const { pathToFileURL } = require('url')` which was added in node 10.12.0.

vweevers: PS. The error message that I'm getting has a more complete explanation than the message shared in the OP. Running `tape test.js` (that's a file on `C:\`) yields:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default
ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

vweevers: @ljharb See https://github.com/nodejs/node/commit/a0846326dd42bf4ec2f91df9cade762283567511. The error message was further tweaked in a later commit, but this should suffice to explain it.

ljharb: @vweevers that makes sense, but I still need to find a way to repro it on windows (but i don't have a windows box) or a test that would have failed on windows (even though we're not running tests on windows yet)

vweevers: Would you take a PR for a basic windows workflow on GitHub Actions (just on latest node to start), with a test? Can split those into 2 PRs if you prefer that

ljharb: Yes, but I prefer not to use the `setup-node` action, and I haven't yet adapted https://github.com/ljharb/actions/tree/main/node/install to work on windows yet.

vweevers: OK, in that case I'll only send a PR for a test (if I can get existing tests to pass on Windows).

ljharb:
@David-Desmaisons thank you so much for bringing the problem to my attention; I'm going to go with the solution in #571, and longer term, we'll look into running CI tests on windows to prevent this kind of problem in the future.

ljharb (approved):

Issue #561, opened 2021-08-31
"Error: The _read() method is not implemented" when used in the browser 


fregante:
I'm unable to run `tape` in the browser with a simple bundle via Parcel. I receive this error at the first test:

```
VM30:1 Uncaught Error: The _read() method is not implemented
    at new o (<anonymous>:8:6099)
    at x.i.register.x._read (<anonymous>:1:10516)
    at x.i.register.x.read (<anonymous>:1:10218)
    at U (<anonymous>:1:8934)
    at x.<anonymous> (<anonymous>:1:11450)
    at x.i.register.a.emit (<anonymous>:1:19390)
    at x.s.resume (<anonymous>:1:3378)
    at x.To.r.resume (<anonymous>:8:84632)
    at x.i.register.x.on (<anonymous>:1:12975)
    at x.i.register.x.pipe (<anonymous>:1:11939)
```

## Repro

1. Create a test.js file:

```js
import test from 'tape';

test('test using promises', t => {
  t.pass();
});
```

2. Run these commands

```sh
echo '{}' > package.json
npm i tape parcel
npx parcel build test.js
```

3. Copy the content of `dist/main.js` and execute it in the browser, like in the console.

## Screenshot

<img width="416" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/131499377-bdaeb7ba-c1b2-48d8-8848-0e2821915161.png">

fregante:
For the record it works if I import https://github.com/martinheidegger/fresh-tape but I'd rather not use forks since `tape`’s line is

> tap-producing test harness for node **and browsers**

ljharb:
Does it work with browserify or webpack < 5?

Parcel might be failing to polyfill node streams.

fregante:
Webpack 4 doesn't bundle at all without custom configuration:
```
❯ webpack ./test.js
Hash: 79ea5ce9a7829ce029a4
Version: webpack 4.46.0
Time: 1569ms
Built at: 09/01/2021 12:54:18 AM
  Asset     Size  Chunks  Chunk Names
main.js  125 KiB       0  main
Entrypoint main = main.js
  [0] ./node_modules/process/browser.js 5.29 KiB {0} [built]
  [1] ./node_modules/call-bind/callBound.js 413 bytes {0} [built]
  [4] ./node_modules/inherits/inherits_browser.js 753 bytes {0} [built]
  [8] ./node_modules/events/events.js 14.5 KiB {0} [built]
[...]

PR #565, opened 2021-09-12
[meta] Exclude `fs` from browser bundlers

fregante:
As suggested in https://github.com/substack/tape/issues/561#issuecomment-909469015 and verified in https://github.com/substack/tape/issues/561#issuecomment-917700972

In short this change will let:

- webpack 4 users bundle without any configuration
- webpack 5 users remove 1 line of config 😅 
- other bundlers to skip looking for the `fs` dependency altogether

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
     "version": "5.3.1",
     "description": "tap-producing test harness for node and browsers",
     "main": "index.js",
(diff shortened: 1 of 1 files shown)


ljharb (approved):

fregante:
Thanks!

Issue #569, opened 2021-11-06
Feature Request: --no-only flag for tape CLI command


EarthyOrange:
tape cli command when invoked with --no-only flag should throw an error if there is a test built with `only` method.

Usage example: `tape test/**/*.js --no-only`

ljharb:
That’s an interesting feature idea. Do any other test runners/frameworks provide it?

ljharb:
I could see something similar with “no todo”, and both being specifiable via an env var, so you can have different values in CI.

EarthyOrange:
@ljharb I haven't used any other test runners/frameworks to be honest. I have however seen cases where developers accidentally commit a test with only method added to it. So the hooks and CI just run that one test and declare success.

ljharb:
Seems reasonable.

jocrah:
hi @ljharb, tape user here. I've been going through the code recently and it has been quite a good learning experience. I would like a few clarifications to help raise a PR for this

jocrah:
> I could see something similar with “no todo”, and both being specifiable via an env var, so you can have different values in CI.

@ljharb does this comment mean we'll be going with the env var approach exclusively instead of a command-line argument as stated in the issue description?

ljharb:
No, both is ideal. However, for each, if an env var is specified a command line flag should overwrite it.

jocrah:
> No, both is ideal/needed. However, for each, if an env var is specified a command line flag should overwrite it.

okay got it, thanks

ljharb:
Fixed by #572.

PR #572, opened 2022-01-22
implement --no-only flag functionality

jocrah:
### Summary
This pr implements the functionality stated in https://github.com/substack/tape/issues/569. 

### How it works
By passing a `--no-only` flag while running tape tests (eg. `tape **/*test.js --no-only`), any `test.only` unintentionally left in any of the tests causes an error to be thrown to make tests fail.

This behaviour can alternatively be achieved by setting the environmental variable `NODE_TAPE_NO_ONLY_TEST` to `true`.


index.js
```diff
@@ -23,9 +24,13 @@ module.exports = (function () {
         wait = true;
     };
 
+    lazyLoad.setCLIOptions = function (options) {
+        if (!options) { options = {}; }
+        cliOptions = options;
+    };
```
ljharb, line 30: i think we should avoid statefulness on the main export. instead, could we have an option passed to `tape.run`, that passes it in turn to `getHarness`?

index.js
```diff
@@ -162,6 +169,7 @@ function createHarness(conf_) {
 
     var only = false;
     test.only = function () {
+        if (conf_.noOnly) { throw new Error('there should be no only test'); }
         if (only) { throw new Error('there can only be one only test'); }
```
ljharb, line 173:
```suggestion
        if (only) { throw new Error('there can only be one `only` test'); }
        if (conf_.noOnly) { throw new Error('`only` tests are disabled'); }
```
let's throw the "bug" one first, and the "disabled" one second

test/no_only.js
```diff
@@ -0,0 +1,120 @@
...
+    { todo: process.versions.node.match(/0\.10\.\d+/) ? 'Fails on node 0.10' : false },
+    function (tt) {
+        tt.plan(2);
+
+        var testStderr = function (rows) {
+            tt.ok((/Error: there should be no only test\n/).test(stripFullStack(rows.toString('utf8')).join('\n')));
+        };
+
+        process.env.NODE_TAPE_NO_ONLY_TEST = true;
+        var ps = spawn(tapeBin, ['**/*.js' ], { cwd: path.join(__dirname, 'no_only') });
```
ljharb, line 41: rather than mutating the env, let's set it in the spawn call.

ljharb (changes requested):
Since test files can be run with either `tape test.js` or `node test.js`, my hope would be that the env var would apply to both cases.

index.js
jocrah: okay will do that

jocrah: @ljharb, so while trying to do this, I realize the `harness` is already defined by the time the `tape.run` is called, hence `getHarness` returns the already created harness which does not have the needed config passed to it.

Kindly let me know if there's something I'm missing. Thanks

ljharb: Hmm, I’ll have to think more about this.

We could make the harness look up the value from the env always, but then we’d have to set it in bin/tape.

jocrah: > We could make the harness look up the value from the env always, but then we’d have to set it in bin/tape.

okay, based on [your comment](https://github.com/substack/tape/pull/572#pullrequestreview-860245563) here, this approach could help achieve that

jocrah: > We could make the harness look up the value from the env always, but then we’d have to set it in bin/tape.

@ljharb  so i've implemented the above suggestion and updated the pr, kindly do let me know if there are no crimes committed while doing so :smile:

index.js
jocrah: done, although I'm thinking `disabled` makes it sound like the `--no-only` flag disables the `.only` tests and would run just fine? 

was thinking maybe `not allowed` or `disallowed`? Do let me know your thoughts on it. Thanks

test/no_only.js
jocrah: done

index.js
ljharb: Good point. Maybe “are prohibited’?

jocrah: okay, sounds good, captures the motive. Making the update then, thanks

jocrah: done

index.js
```diff
@@ -66,7 +66,8 @@ module.exports = (function () {
 function createExitHarness(conf, wait) {
     var config = conf || {};
     var harness = createHarness({
-        autoclose: defined(config.autoclose, false)
+        autoclose: defined(config.autoclose, false),
+        noOnly: defined(process.env.NODE_TAPE_NO_ONLY_TEST, false)
```
ljharb, line 70: let's allow this to be passed in with `conf`
```suggestion
        noOnly: defined(conf.noOnly, defined(process.env.NODE_TAPE_NO_ONLY_TEST, false))
```

readme.markdown
```diff
@@ -144,6 +144,23 @@ By default, uncaught exceptions in your tests will not be intercepted, and will
...
+## --no-only
+This is particularly useful in a CI environment where an [only test](#testonlyname-opts-cb) is not supposed to go unnoticed.
+
+By passing the `--no-only` flag, any existing [only test](#testonlyname-opts-cb) causes tests to fail.
+
+```sh
+tape **/*.js --no-only
+```
+
+Alternatively, environmental variable `NODE_TAPE_NO_ONLY_TEST` can be set to `true` to achieve same behavior. Do note that the command-line flag takes precedence.
```
ljharb, line 162:
```suggestion
Alternatively, the environment variable `NODE_TAPE_NO_ONLY_TEST` can be set to `true` to achieve the same behavior; the command-line flag takes precedence.
```

readme.markdown
```diff
@@ -144,6 +144,23 @@ By default, uncaught exceptions in your tests will not be intercepted, and will
 - In-process reporting with https://github.com/DavidAnson/tape-player
 - Describe blocks with https://github.com/mattriley/tape-describe
 
+# command-line flags
+
+While running tests, top-level configurations can be passed via the command line to specify desired behavior.
```
ljharb, line 149: if we're going to make this section, then `--require` and `--import` need to be documented also.

readme.markdown
```diff
@@ -144,6 +144,23 @@ By default, uncaught exceptions in your tests will not be intercepted, and will
...
+
+Available configurations are listed below:
+
+## --no-only
+This is particularly useful in a CI environment where an [only test](#testonlyname-opts-cb) is not supposed to go unnoticed.
+
+By passing the `--no-only` flag, any existing [only test](#testonlyname-opts-cb) causes tests to fail.
```
ljharb, line 156: let's link to this section from the `test.only` section.

readme.markdown
```diff
@@ -144,6 +144,23 @@ By default, uncaught exceptions in your tests will not be intercepted, and will
...
+
+```sh
+tape **/*.js --no-only
```
ljharb, line 159:
```suggestion
tape --no-only **/*.js
```
generally, positional arguments must come after dashed arguments.

test/no_only.js
```diff
@@ -0,0 +1,121 @@
...
+var spawn = require('child_process').spawn;
+var concat = require('concat-stream');
+
+var stripFullStack = require('./common').stripFullStack;
+
+var tapeBin = path.join(process.cwd(), 'bin/tape');
+
+tap.test(
+    'Should throw error when --no-only is passed via cli and there is a .only test',
+    { todo: process.versions.node.match(/0\.10\.\d+/) ? 'Fails on node 0.10' : false },
```
ljharb, line 14: why does it fail on node 0.10?

jocrah: okay, so for node 0.10, the exit code was 8 instead of 1 for the cases where the error is thrown. 

I tried to look for similar use-cases in the tests and saw we try to skip such behaviour by excluding some versions, notable amongst them was https://github.com/substack/tape/blob/master/test/async-await.js.

ljharb: That test file skips in node < 8 because node 8 is what adds `async`/`await` syntax.

I can't find anywhere exiting with an `8`; I'll take a look at the failures and see what might be the problem.

jocrah: o okay makes sense. I installed 0.10 locally, removed the condition and run the tests and got the code as `8` for that as well hence my conclusion. I get the reason now though.

readme.markdown
jocrah: okay, just to clarify, I rather see `--require` and `--ignore` in the code? Kindly let me know if there's more to it.

index.js
jocrah: updated

readme.markdown
jocrah: updated

readme.markdown
ljharb: oh right, sorry, yes, those are the two.

jocrah: updated pr to add documentation for `--require` and `--ignore` so far

readme.markdown
jocrah: done

readme.markdown
jocrah: updated

jocrah:
> I tried to improve this by passing an option explicitly into `tape.run`, but because of the way the lazy loader works, i'm not sure how well that'd work.

o okay yea understood, also been trying to debug the exit code being `8` for the node version 0.10, read around it and got to know some earlier versions of node used it for uncaught exceptions but not sure what can be done yet at our end

ljharb:
ah right, forgot about that. i'll take a quick look locally.

ljharb:
I see that 0.10 and 0.8 both have different kinds of failures, including some exit codes of 127. io.js v3.0-v3.2 seem to have similar failures. I'd prefer not to start making these fail, so let's look more into them.

I've rebased on top of a commit that fixes the 16.9/16.10 failures.

jocrah:
> I see that 0.10 and 0.8 both have different kinds of failures, including some exit codes of 127. io.js v3.0-v3.2 seem to have similar failures. I'd prefer not to start making these fail, so let's look more into them.

okay got it, thanks. Earlier today, tried 0.8 locally and did not get the tests introduced in this pr failing (I guess could depend on the specific minor version). 

Please do you mean some other tests failed also with the 127 exit code or these same tests?

ljharb:
At this point (and we can check once CI is completed) i think all the remaining failures are:
 - 0.6 (and i think 0.9), which always fail
 - node 0.8, 0.10, 3.0, 3.1, and 3.2, failing on `test/no_only.js`
 
 The latter is my concern.

jocrah:
so after trying locally, I see the iojs 3.0, 3.1 and 3.2 failures are related to some stack-buffer overflow error related to https://github.com/nodejs/node/issues/2581, and fixed i think for latter versions?

ljharb:
hmm, meaning what, that this one test file is triggering a bug, but no other test files have triggered it before?

ljharb:
hmm, oddly enough locally v3.3 fails, but it passes in CI.

ljharb:
hmm, when i run the test file by itself, it fails consistently for me locally, but when i run the full suite, it passes. That's concerning.

ljharb:
ok, figured that last one out; when specifying `env`, the PATH has to be explicitly passed down.

ljharb:
(also, tap v8 doesn't seem to have `todo`, which is added in v11)

test/no_only.js
```diff
@@ -0,0 +1,123 @@
...
+    { todo: expectedFailure ? 'Fails on these node versions' : false },
+    function (tt) {
+        tt.plan(2);
+
+        var testStderr = function (rows) {
+            tt.ok((/Error: `only` tests are prohibited\n/).test(stripFullStack(rows.toString('utf8')).join('\n')));
+        };
+        var ps = spawn(tapeBin, ['**/*.js'], {
+            cwd: path.join(__dirname, 'no_only'),
+            env: { PATH: process.env.PATH, NODE_TAPE_NO_ONLY_TEST: false }
```
jocrah, line 43: for this, the test case was initially meant to test out behaviour when `NODE_TAPE_NO_ONLY_TEST` is passed as `true` and there is an existing `only` test. A bit confused about the change to `false`?

ljharb: oh oops! Just a bad copy-paste. Will fix.

jocrah: o okay cool, but anyway, was wondering why the related tests passed anyway, tried locally and realized the env is evaluated as a string in the code so always evaluates to `true`

ljharb:
When I use iojs v3.2, and I run `tape --no-only test-b.js`, i get an exit code of 1 and the expected output - is the bug just around spawn?

jocrah:
o okay interesting, yea most likely something around that.

ljharb:
I've switched from spawn to exec; that seems like it'll fix it :-)

ljharb:
yeah, i locally changed the `ok` to a `match`, and i see it's the exact node bug you linked.

ljharb:
I went ahead and TODOd the tests for node 0.8, 0.10, and 3.0-3.2.

jocrah:
okay cool, this was fun :smile:, thanks for the review and insights @ljharb

index.js
jocrah: just out of curiosity and to help improve, maybe one thing left lingering would be to appreciate why statefulness on the main export would be good to avoid as highlighted [here](https://github.com/substack/tape/pull/572#discussion_r790189739). if that's okay to ask

PR #576, opened 2022-05-27
stream locking

### LINKED ISSUE #575: Cannot create a stream then load ESM test modules
Hi, the following won't work :

```js
import test from 'tape';

test.createStream({ objectMode: true }).on('data', function (row) {
    console.log(JSON.stringify(row));
});

import('./testESMModule.mjs');
```

```js
/* ./testESMModule.mjs */

import test from 'tape';

test('This one should pass', function (t) {
    t.pass();
    t.end();
});
```
Actually the test in the ESM Module does not run because the stream creates a result queue that is exhausted in the same event loop it is created in (result are consumed in a `setImmediate` callback). The ESM module will be loaded asynchronously, in a subsequent event loop.  

The `wait` and `run` function used in the cli are not helpful here because the stream will be recreated on the `run` method, thus :
 - the stream will not be the one with registered listeners.
 - the tests are not registered on the stream because the `run` method is called after the test are defined in the test module.

For now I am using this trick :
```js
import test from 'tape';

test.createStream({ objectMode: true }).on('data', function (row) {
    console.log(JSON.stringify(row));
});

test('__load_async__', async function (t) {
    await import('./testESMModule.mjs');
    t.end();
});
```
The result processing starts by the synchronously defined test named `'__load_async__'` in the same event loop in which the stream is created. It will then "lock" the stream until all modules are loaded (because `t.end()` is called once they're all loaded). Then the results processing continues with the asynchronously defined tests.

I propose something better, which implementation uses the same kind of mechanism as above : 
```js
import test from 'tape';

test.createStream({ objectMode: true }).on('data', function (row) {
    console.log(JSON.stringify(row));
});

test.async(async function () {
    await import('./testESMModule.mjs');
});
```
I implemented this here : https://github.com/ydarma/tape/blob/async-load/test/async-test-load.js (I can make a PR).

ydarma:
This fixes the `wait`, `run` calls to enable use with object mode streams and asynchronous module loading.

fix #575 close #575


index.js
```diff
@@ -142,10 +147,15 @@ function createExitHarness(conf, wait) {
 		stream.on('end', function () { ended = true; });
 	}
 
+	run();
+
 	if (wait) {
```
ljharb, line 152: does this change mean it's breaking for anyone using `wait`?

PR #584, opened 2022-11-16
remove has dependency

Uzlopak:


ljharb (changes requested):
Why? `has` is a very commonly used dep that does the right thing.

ljharb:
Note that fewer dependencies is never inherently better.

Uzlopak:
Well if you check it is basically just a Object.prototype.hasOwnProperty wrapper without any polyfill/ponyfill/shim.

Uzlopak:
Closing because i have already over 800+ open PRs. No interest in filling the list of open PRs further.

PR #587, opened 2022-11-24
[New] Add `--ignore-pattern` flag 

ppati000:
This PR introduces the --ignorePattern flag as a shorthand for ignoring test files.

The flag may be used together with --ignore; the input will be concatenated in that case. I figured this behavior would make the most sense, since users may want to ignore everything in .gitignore and some other files on top.

Let me know if this looks good, then I'll also add some documentation :)

Fixes https://github.com/ljharb/tape/issues/586

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -7,9 +7,9 @@ var objectKeys = require('object-keys');
 
 var opts = parseOpts(process.argv.slice(2), {
 	alias: { r: 'require', i: 'ignore' },
-	string: ['require', 'ignore'],
+	string: ['require', 'ignore', 'ignore-pattern'],
 	boolean: ['only'],
-	default: { r: [], i: null, only: null }
+	default: { r: [], i: null, 'ignore-pattern': null, only: null }
 });
 
 if (typeof opts.only === 'boolean') {
@@ -35,15 +35,23 @@ opts.require.forEach(function (module) {
 var resolvePath = require('path').resolve;
 var requireResolve = require.resolve;
 
-var matcher;
+var ignoreStr = '';
 if (typeof opts.ignore === 'string') {
 	var readFileSync = require('fs').readFileSync;
 	try {
-		var ignoreStr = readFileSync(resolvePath(cwd, opts.ignore || '.gitignore'), 'utf-8');
+		ignoreStr = readFileSync(resolvePath(cwd, opts.ignore || '.gitignore'), 'utf-8');
 	} catch (e) {
 		console.error(e.message);
 		process.exit(2);
 	}
+}
+
+if (typeof opts['ignore-pattern'] === 'string') {
+	ignoreStr += '\n' + opts['ignore-pattern'];
+}
+
+var matcher;
+if (ignoreStr) {
 	var ignore = require('dotignore');
 	matcher = ignore.createMatcher(ignoreStr);
 }
--- a/readme.markdown
+++ b/readme.markdown
@@ -158,13 +158,24 @@ This is used to load modules before running tests and is explained extensively i
 
 **Alias**: `-i`
 
-This flag is used when tests from certain folders and/or files are not intended to be run. It defaults to `.gitignore` file when passed with no argument.
+This flag is used when tests from certain folders and/or files are not intended to be run.
+The argument is a path to a file that contains the patterns to be ignored.
+It defaults to `.gitignore` when passed with no argument.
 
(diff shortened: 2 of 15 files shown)


scottk (approved):

david-gregorian (approved):

bin/tape
```diff
@@ -36,14 +36,24 @@ var resolvePath = require('path').resolve;
...
 var matcher;
-if (typeof opts.ignore === 'string') {
-	var readFileSync = require('fs').readFileSync;
-	try {
-		var ignoreStr = readFileSync(resolvePath(cwd, opts.ignore || '.gitignore'), 'utf-8');
-	} catch (e) {
-		console.error(e.message);
-		process.exit(2);
+var isIgnoreFlag = typeof opts.ignore === 'string'
+var isIgnorePatternFlag = typeof opts.ignorePattern === 'string'
```
ljharb, line 40:
```suggestion
var isIgnoreFlag = typeof opts.ignore === 'string';
var isIgnorePatternFlag = typeof opts.ignorePattern === 'string';
```

bin/tape
```diff
@@ -36,14 +36,24 @@ var resolvePath = require('path').resolve;
...
+
+if (isIgnoreFlag || isIgnorePatternFlag) {
+	var ignoreStr = ''
```
ljharb, line 43:
```suggestion
	var ignoreStr = '';
```

bin/tape
```diff
@@ -36,14 +36,24 @@ var resolvePath = require('path').resolve;
...
+		var readFileSync = require('fs').readFileSync;
+		try {
+			ignoreStr = readFileSync(resolvePath(cwd, opts.ignore || '.gitignore'), 'utf-8');
+		} catch (e) {
+			console.error(e.message);
+			process.exit(2);
+		}
+	}
+	if (isIgnorePatternFlag) {
+		ignoreStr = ignoreStr.concat("\n" + opts.ignorePattern)
```
ljharb, line 54:
```suggestion
		ignoreStr += "\n" + opts.ignorePattern;
```

test/ignore_pattern.js
```diff
@@ -0,0 +1,34 @@
...
+
+var tap = require('tap');
+var path = require('path');
+var execFile = require('child_process').execFile;
+
+var tapeBin = path.join(process.cwd(), 'bin/tape');
+
+tap.test('should allow ignore file together with --ignorePattern', function (tt) {
+	tt.plan(1);
+	var proc = execFile(tapeBin, ['**/*.js', '--ignore', '.ignore', '--ignorePattern', 'fake_other_ignored_dir'], { cwd: path.join(__dirname, 'ignorePattern') });
```
ljharb, line 11: i'd expect the file paths to come last, after all the dashed arguments. any reason to order them this way?

ljharb (changes requested):
Seems pretty reasonable, assuming the tests fail without the bin/tape changes.

test/ignore_pattern.js
ppati000: No particular reason, just changed the order 👌

ljharb:
I've rebased this, and renamed the flag to `ignore-pattern` to more closely match command-line convention.

All that's missing now is documentation.

ppati000:
@ljharb cool, thank you 👌 I've added some documentation now, feel free to make any adjustments as needed!

readme.markdown
```diff
@@ -158,14 +158,31 @@ This is used to load modules before running tests and is explained extensively i
...
+## --ignore-pattern
+
+Same functionality as `--ignore`, but passing the pattern directly instead of an ignore file.
+If both `--ignore` and `--ignore-pattern` are given, the `--ignore-pattern` argument is appended to the content of the ignore file.
+
+```sh
+tape --ignore-pattern integration_tests/**/*.js **/*.js
+```
+
+**Note:** Your shell might auto-expand the provided arguments. In this case, quote the arguments like this:
```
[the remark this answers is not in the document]
ppati000: Probably not in the default configuration. You can probably disable wildcards in most shells though, see for example [here](https://stackoverflow.com/a/11456453). Probably not worth a mention though, we can just go with quoted examples imo.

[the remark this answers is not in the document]
ljharb: Sounds good, let's do that.

ljharb (approved):

PR #596, opened 2023-06-15
Link to explain what TAP is

michaelkirk:
I spent a few minutes trying to understand what "tap-producing" meant. Maybe I'm not the only one?

Also, I capitalized it to help ignorant people like me infer it's an acronym.


ljharb (approved):
Sounds good, thanks

PR #183, opened 2015-08-27
Fix parsing of stack traces on windows

Joris-van-der-Wel:
`functionName`, `file`, `line`, `column` will now be properly set in result objects (and subsequently tape text output).

I wanted to add a test case, however having to workaround `__dirname` is probably not worth the effort (and would make the test case itself change behaviour based on the host OS too)

PR #602, opened 2024-01-09
Fix path parsing and tests on windows

Joris-van-der-Wel:
This is a follow up of #183 which involved a similar fix for a much older version (9 years ago) of this library. I noted 4 years ago that the stack trace parsing on windows had already been fixed by someone else. 

However I just discovered that the issue has cropped up again, which this PR attempts to fix. Luckily this time the existing tests already check for this issue, however the tests must be ran on windows. This uncovered more issues because a lot of tests were broken on windows.

Some tests were attempting to spawn `bin/tape`, instead of `node bin/tape`. This does not work on my windows machine.

The failing tests uncovered that the `--ignore` and `--ignore-pattern` options do not work properly on windows. This was because the old version of glob uses forward slashes on windows, while dotignore expects backslashes.

It was not possible to run the tests using `npm test`. This was because of a bug in the old versions of `nyc`, and also because single quotes were used in the run script.

Before:

```
  303 passing (14s)
  8 pending
  64 failing
```

After:

```
  374 passing (31s)
  5 pending
```

The individual commits in this PR contain more details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
 		"js-yaml": "^3.14.0",
 		"npm-run-posix-or-windows": "^2.0.2",
 		"npmignore": "^0.3.0",
+		"nyc": "^10.3.2",
 		"safe-publish-latest": "^2.0.0",
 		"tap": "^8.0.1",
 		"tap-parser": "^5.4.0"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
 		"dotignore": "^0.1.2",
 		"for-each": "^0.3.3",
 		"get-package-type": "^0.1.0",
-		"glob": "^7.2.3",
+		"glob": "^10.3.10",
```
ljharb, line 37: we can't upgrade `glob` past v7, because v8 drops support for node < 12.

package.json
```diff
@@ -64,6 +64,7 @@
 		"js-yaml": "^3.14.0",
 		"npm-run-posix-or-windows": "^2.0.2",
 		"npmignore": "^0.3.0",
+		"nyc": "^15.1.0",
```
ljharb, line 67: Similarly, we can't upgrade nyc past v10.

ljharb (changes requested):
Thanks, this is amazing!

Joris-van-der-Wel:
I have a new version of these commits without any dependency update. They are on a [different branch](https://github.com/ljharb/tape/compare/master...Joris-van-der-Wel:tape:windows2).

How would you prefer to receive these new changes? Force push? New PR? Extra commit?

ljharb:
Force push to this branch would be great :-) thanks!

ljharb (approved):
Love it, thank you! I'll verify tests locally and in a windows VM before merging.

PR #604, opened 2024-03-07
[Dev Deps] remove unused `intl-fallback-symbol`, `is-core-module`

Uzlopak:
is-core-module is nowhere used


ljharb (approved):
Thanks! these are dev deps, so it's not a big deal, but nice to clean it up.

sadly `depcheck` doesn't check extensionless files, so it's not a reliable tool in this repo.

Combined with #606.

PR #605, opened 2024-03-07
chore: remove jackspeak

Uzlopak:
jackspeak is nowhere used


ljharb (changes requested):
This is definitely used, just transitively, and is pinned per a576f8d

PR #606, opened 2024-03-07
[Dev Deps] remove unused `intl-fallback-symbol`, `is-core-module`

Uzlopak:
intl-fallback-symbol is nowhere used


ljharb:
Combined with #604

PR #621, opened 2026-02-04
[Refactor] Use `globv7` instead of old glob 7.2.3

xsfj:
The glob package (v7.2.3) currently used is deprecated. While it functions, it is no longer receiving security patches or bug fixes from the original author, who has moved on to v8+.

Moving tape to glob v8/v9/v10 would require a significant refactor because those versions are now promise-based and have different API signatures, and it would also break backwards compatibility.

Solution: I've switched the dependency to globv7. This is a maintained fork of the original glob v7.2.3 logic. It’s a drop-in replacement that preserves the synchronous/callback behavior tape relies on. It provides a path for security patches if vulnerabilities are found in the v7 logic without forcing a breaking change on this repository.

My fork maintains the same backwards compatibility and API

I did not use AI to generate any part of the code for this change.

### DIFF
--- a/bin/tape
+++ b/bin/tape
@@ -57,7 +57,7 @@ if (ignoreStr) {
 	matcher = ignore.createMatcher(ignoreStr);
 }
 
-var glob = require('glob');
+var glob = require('globv7');
 
 var files = opts._.reduce(function (result, arg) {
 	if (glob.hasMagic(arg)) {
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
 		"dotignore": "^0.1.2",
 		"for-each": "^0.3.5",
 		"get-package-type": "^0.1.0",
-		"glob": "^7.2.3",
+		"globv7": "^1.0.0",
 		"has-dynamic-import": "^2.1.1",
 		"hasown": "^2.0.2",
 		"inherits": "^2.0.4",


ljharb:
I am absolutely not going to switch from a widely used package to an unknown unadopted fork.

There remains a path for security fixes to go into v7; the author is responsive and his willingness is not knowable in advance.

Separately, there's nothing inherently wrong with depending one something that might not receive security patches, _until a security vulnerability exists_.

xsfj:
I am not saying that a deprecation warning is a bug. I am only saying that if there is a vulnerability, Glob version 7 will not be updated, as the author has said so. Even if it is unknown, if you read the code, it's literally exactly what I said.

ljharb:
Yes, I realize that. That's not a problem worth solving until a vulnerability exists, especially given that ANY potential vulnerability in glob would be in the `tape` executable you're running yourself, which would mean that it'd be a self-attack - iow, i can't even conceive of a vulnerability in glob that would end up being an actual vulnerability in tape's usage of glob.

xsfj:
> There remains a path for security fixes to go into v7; the author is responsive and his willingness is not knowable in advance.

[As quoted by the author](https://www.npmjs.com/package/glob/v/7.2.3):

> Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me."

I do not believe he has any intention of patching security vulnerabilities for versions he considers obsolete.

In [CVE-2025-64756](https://nvd.nist.gov/vuln/detail/CVE-2025-64756), it took an entire community of mad developers to make the author make a patch for v10. If he was very unwilling to patch the previous major version, why would he patch v7?

xsfj:
I understand your hesitancy to use an unpopular package from an unknown developer in your testing framework, but depending on old versions of libraries that the author has explicitly stated are abandoned is like building your stack on loose Jenga blocks. If one falls, the whole thing falls.

ljharb:
As I explained, though, the risk is effectively zero no matter what vulnerabilities glob might have.

xsfj:
> Separately, there's nothing inherently wrong with depending one something that might not receive security patches, _until a security vulnerability exists_.

Now a vulnerability was found in minimatch that affects glob. I still think if you don't want to use mine, just make your own fork of it.

ljharb:
minimatch will be receiving backports to fix that vulnerability, so there remains no issue.

PR #624, opened 2026-06-06
[Fix] Blocking tests exceeding the test timeout fail

### LINKED ISSUE #623: Blocking tests don't cause tests to fail when exceeding the timeout time limit
When a test is slow in a blocking fashion, tape's timeout does not apply. A blocking test can take 10s and pass with a 5s timeout. The following test suite demonstrates the asymmetry between blocking and non-blocking tests (with a 10ms runtime and 5ms timeout instead):

```javascript
var test = require('tape');

test('opts.timeout, blocking', { timeout: 5 }, function (t) {
	var start = Date.now();
	while (Date.now() < start + 10) { /* Busy-wait to block the event loop */ }
	t.pass("slow success");
	t.end();
});

test('timeoutAfter, blocking', function (t) {
	t.timeoutAfter(5);
	var start = Date.now();
	while (Date.now() < start + 10) { /* Busy-wait to block the event loop */ }
	t.pass("slow success");
	t.end();
});

test('opts.timeout, async', { timeout: 5 }, function () {
	setTimeout(function () { }, 10);
});

test('timeoutAfter, async', function (t) {
	t.timeoutAfter(5);
	setTimeout(function () { }, 10);
});
```

If you run it (using an up-to-date version of Node.js and tape) the first two tests will pass and the other two will fail:

```shell
$ node --version
v26.0.0

$ npm ls tape | grep tape
-- tape@5.9.0

$ node minimal-repro.js 
TAP version 13
# opts.timeout, blocking
ok 1 slow success
# timeoutAfter, blocking
ok 2 slow success
# opts.timeout, async
not ok 3 opts.timeout, async timed out after 5ms
  ---
    operator: fail
    stack: |-
      Error: opts.timeout, async timed out after 5ms
          at Test.assert [as _assert] (/home/eric/workspace/node-assert-time/node_modules/tape/lib/test.js:492:48)
          at Test.fail (/home/eric/workspace/node-assert-time/node_modules/tape/lib/test.js:600:7)
          at Timeout._onTimeout (/home/eric/workspace/node-assert-time/node_modules/tape/lib/test.js:194:8)
          at listOnTimeout (node:internal/timers:605:17)
          at process.processTimers (node:internal/timers:541:7)
  ...
# timeoutAfter, async
not ok 4 timeoutAfter, async timed out after 5ms
  ---
    operator: fail
    stack: |-
      Error: timeoutAfter, async timed out after 5ms
          at Test.assert [as _assert] (/home/eric/workspace/node-assert-time/node_modules/tape/lib/test.js:492:48)
          at Test.fail (/home/eric/workspace/node-assert-time/node_modules/tape/lib/test.js:600:7)
          at Timeout._onTimeout (/home/eric/workspace/node-assert-time/node_modules/tape/lib/test.js:194:8)
          at listOnTimeout (node:internal/timers:605:17)
          at process.processTimers (node:internal/timers:541:7)
  ...

1..4
# tests 4
# pass  2
# fail  2
```

I would expect all these tests to fail. As a point of comparison [Mocha](https://mochajs.org/) does behave this way ([Mocha PoC](https://github.com/ericcornelissen/node-assert-time/blob/55d82ffb47cbcc70b521921f0794265a2c389294/example-mocha.js#L8-L56)) whereas [AVA](https://github.com/avajs/ava) has the same limitation as tape ([AVA PoC](https://github.com/ericcornelissen/node-assert-time/blob/55d82ffb47cbcc70b521921f0794265a2c389294/example-ava.js#L8-L52)).

If tape should indeed fail in all four cases, I believe the fix implementation can be based off of [my `assert-time` package](https://github.com/ericcornelissen/node-assert-time/blob/55d82ffb47cbcc70b521921f0794265a2c389294/index.js) (which is also MIT licensed, so the code can easily be used for tape).

I (somewhat surprisingly) did not find an existing issue for this problem. The closest issues seem to be #161 (just about displaying times) and #188 (which is also about the behavior of timeouts, but does not mention blocking tests).

What caused me to create this issue is https://github.com/es-shims/String.prototype.trim/pull/13 (and transitively to related Pull Requests from me in various other, similar projects). If tape behaved "properly" for long-running blocking tests, that (and the related) contribution would be simpler.

ericcornelissen:
Closes #623

Change the [`timeoutAfter`](https://github.com/tape-testing/tape/blob/39470e7d6adea7f44b3fbe380e242298b012b3b2/lib/test.js#L189) method to _also_ fail if a blocking test's runtime duration exceeds the timeout. Doing it here ensures there's one location in the code that handles both the `timeout` option and `timeoutAfter` assertion. Use a boolean flag `timedOut` to avoid double failures in the event of a race condition.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -189,12 +189,24 @@ Test.prototype.plan = function plan(n) {
 Test.prototype.timeoutAfter = function timeoutAfter(ms) {
 	if (!ms) { throw new Error('timeoutAfter requires a timespan'); }
 	var self = this;
+	var startTime = Date.now();
+	var timedOut = false;
 	var timeout = safeSetTimeout(function () {
+		if (timedOut) { return; }
+		timedOut = true;
 		self.fail(self.name + ' timed out after ' + ms + 'ms');
 		self.end();
 	}, ms);
 	this.once('end', function () {
 		safeClearTimeout(timeout);
+	});
+	this.once('end', function () {
+		if (timedOut) { return; }
+		timedOut = true;
+		var duration = Date.now() - startTime;
+		if (duration > ms) {
+			self.fail(self.name + ' timed out after ' + duration + 'ms');
+		}
 	});
 };
 
--- a/test/timeout.js
+++ b/test/timeout.js
@@ -1,17 +1,65 @@
 'use strict';
 
-var test = require('../');
+var tape = require('../');
+var tap = require('tap');
+var concat = require('concat-stream');
+
+var stripFullStack = require('./common').stripFullStack;
+
 var ran = 0;
 
-test('timeout', function (t) {
-	t.pass('this should run');
-	ran++;
-	setTimeout(function () {
+tap.test('timeout with multiple tests', function (tt) {
+	tt.plan(0);
+
+	var test = tape.createHarness();
+
+	test('timeout', function (t) {
+		t.pass('this should run');
+		ran++;
+		setTimeout(function () {
+			t.end();
+		}, 100);
+	});
+
+	test('should still run', { timeout: 50 }, function (t) {
+		t.equal(ran, 1);
 		t.end();
-	}, 100);
+	});
 });
 
-test('should still run', { timeout: 50 }, function (t) {
-	t.equal(ran, 1);
-	t.end();
+tap.test('timeout with blocking', function (tt) {
+	tt.plan(1);
+
+	var test = tape.createHarness();
+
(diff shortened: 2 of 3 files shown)


ljharb (changes requested):
This would be a breaking change. I've got a local change that adds an option to have this behavior, that we can change in a future semver-major.

ericcornelissen:
Works for me

